PR #2, opened 2016-04-08
TmpDirStep appears never to have been used

jglick:
Introduced by @kohsuke in https://github.com/jenkinsci/pipeline-plugin/commit/372efa4618d711f2ab8b778bf5e5388a2956833b.

@reviewbybees


amuniz:
How do you know it's not used?

jglick:
> How do you know it's not used?

Because I [looked](https://github.com/search?utf8=%E2%9C%93&q=user%3Ajenkinsci+user%3Acloudbees+TmpDirStep&type=Code&ref=searchresults). Anyway this was only a test class, so it could not have been used from main code.

amuniz:
:bee:

PR #4, opened 2016-06-09
[JENKINS-26130] Way to print progress from pickles

jglick:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/5.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/TryRepeatedly.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/TryRepeatedly.java
@@ -26,11 +26,17 @@ package org.jenkinsci.plugins.workflow.support.pickles;
 
 import com.google.common.util.concurrent.AbstractFuture;
 import com.google.common.util.concurrent.ListenableFuture;
+import hudson.model.TaskListener;
+import java.io.IOException;
 import jenkins.util.Timer;
 
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 
 /**
  * {@link ListenableFuture} that promises a value that needs to be periodically tried.
@@ -38,6 +44,9 @@ import javax.annotation.CheckForNull;
  * @author Kohsuke Kawaguchi
  */
 public abstract class TryRepeatedly<V> extends AbstractFuture<V> {
+
+    private static final Logger LOGGER = Logger.getLogger(TryRepeatedly.class.getName());
+
     private final int delay;
     private ScheduledFuture<?> next;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/pickles/TryRepeatedly.java
```diff
@@ -26,20 +26,36 @@
...
 public abstract class TryRepeatedly<V> extends AbstractFuture<V> {
+
+    private static final Logger LOGGER = Logger.getLogger(TryRepeatedly.class.getName());
+
     private final int delay;
     private ScheduledFuture<?> next;
+    /** Number of {@link #tryLater} calls to run between logging attempts. */
+    private float backoff = 1;
+    /** Amount by which {@link #backoff} gets multiplied, so we do not flood the log with endless messages. */
+    private static final float BACKOFF_EXPONENT = 1.42f; // >√2
```
oleg-nenashev: You could use the Golden ratio constant to make the code more scientific :)

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/TryRepeatedly.java
```diff
@@ -73,8 +114,10 @@ public void run() {
 
     @Override
     public boolean cancel(boolean mayInterruptIfRunning) {
-        if (next!=null)
+        if (next != null) {
             next.cancel(mayInterruptIfRunning);
+        }
+        LOGGER.fine("cancelling " + this + " in " + getOwner());
```
oleg-nenashev: uppercase?

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/TryRepeatedly.java
jglick: :-)

The exponent here is pretty arbitrary, I just chose something which seemed to behave reasonably by watching it interactively.

`DurableTaskStep.Execution` uses a lower exponent, but there we want to actually grab process output sooner rather than later, whereas here the message is usually redundant, we just want to make it clear in the build log that something is still happening.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/PickleResolver.java
```diff
@@ -49,15 +51,28 @@
...
+        this.owner = owner;
     }
 
     public Object get(int id) {
         return values.get(id);
     }
 
+    @Deprecated
     public ListenableFuture<PickleResolver> rehydrate() {
+        return rehydrate(new ArrayList<ListenableFuture<?>>());
```
oleg-nenashev: Why don't you just use the same implementation in https://github.com/jenkinsci/workflow-api-plugin/pull/5?

jglick: Not sure what you mean. This is just a compatibility stub in case a user neglects to update `workflow-cps` to match.

oleg-nenashev: Uh ok

oleg-nenashev:
🐝

PR #5, opened 2016-06-28
[JENKINS-30412] Offering a sandbox-friendly RunWrapper.changeSets property

jglick:
[JENKINS-30412](https://issues.jenkins-ci.org/browse/JENKINS-30412)

Downstream of https://github.com/jenkinsci/script-security-plugin/pull/78.

Also needs https://github.com/jenkinsci/pipeline-build-step-plugin/pull/4 for documentation; I struggled to find a way for `BuildTriggerStep/help-wait.html` and `RunWrapperBinder/help.jelly` to `st:include` some common file from `workflow-support`, but could not make it work. Specifiying just `path` throws an NPE from `IncludeTag`; you need to specify also `class`, but this is then not found since the class is in a plugin loader, not core. Not sure how [this](https://github.com/jenkinsci/groovy-postbuild-plugin/blob/9f49c9e128c6124fd01d44239565e5eb5388ba54/src/main/resources/org/jvnet/hudson/plugins/groovypostbuild/WorkflowManager/help.jelly#L28) works. Could probably switch to Groovy views and do it somehow. Anyway, a topic for a future refactoring.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,8 +28,8 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.5</version>
-        <relativePath />
+        <version>2.11</version>
+        <relativePath/>
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-support</artifactId>
@@ -52,17 +52,18 @@
     <repositories>
         <repository>
             <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
+            <url>https://repo.jenkins-ci.org/public/</url>
         </repository>
     </repositories>
     <pluginRepositories>
         <pluginRepository>
             <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
+            <url>https://repo.jenkins-ci.org/public/</url>
         </pluginRepository>
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
+        <jenkins-test-harness.version>2.11</jenkins-test-harness.version>
     </properties>
     <dependencies>
         <dependency>
@@ -78,7 +79,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
-            <version>1.16</version>
+            <version>1.21-SNAPSHOT</version>
         </dependency>
         <dependency>
             <groupId>org.jboss.marshalling</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -52,17 +52,18 @@
...
     <pluginRepositories>
         <pluginRepository>
             <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
+            <url>https://repo.jenkins-ci.org/public/</url>
         </pluginRepository>
     </pluginRepositories>
     <properties>
         <jenkins.version>1.642.3</jenkins.version>
+        <jenkins-test-harness.version>2.11</jenkins-test-harness.version>
```
jglick: Need `getLog` fixed to be plain text.

pom.xml
```diff
@@ -128,6 +129,26 @@
...
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>git</artifactId>
+            <version>2.5.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>git</artifactId>
+            <version>2.5.0</version>
+            <classifier>tests</classifier>
```
jglick: `GitSampleRepoRule`

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -177,4 +179,14 @@ public String getId() throws AbortException {
         return build().getAbsoluteUrl();
     }
 
+    @Whitelisted
+    public List<ChangeLogSet<? extends ChangeLogSet.Entry>> getChangeSets() throws Exception {
+        Run<?,?> build = build();
+        try { // TODO JENKINS-24141 should not need to use reflection here
+            return (List) build.getClass().getMethod("getChangeSets").invoke(build);
```
jglick: currently this will work on `AbstractBuild` and `WorkflowRun`

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapperTest.java
```diff
@@ -99,4 +106,58 @@
...
+                sampleRepo2.git("add", "elsewhere");
+                sampleRepo2.git("commit", "--message=second repo");
+                assertThat(JenkinsRule.getLog(r.j.assertBuildStatusSuccess(p.scheduleBuild2(0))), containsRegexp(
+                    "changeSets: kind=git; entries=[a-f0-9]{40} by .+ ~ .+ on .+: more stuff: (edit onefile; add anotherfile|add anotherfile; edit onefile), " +
+                    "[a-f0-9]{40} by .+ ~ .+ on .+: amended: edit onefile & " +
+                    "kind=git; entries=[a-f0-9]{40} by .+ ~ .+ on .+: second repo: add elsewhere"));
+            }
+        });
+    }
+    // TODO why is this not in Hamcrest to begin with?
```
jglick: Hmm, seems to be in https://github.com/hamcrest/JavaHamcrest/pull/57 but where is that?

jglick: Found it in

```xml
<dependency>
    <groupId>org.hamcrest</groupId>
    <artifactId>java-hamcrest</artifactId>
    <version>2.0.0.0</version>
    <scope>test</scope>
</dependency>
```

but less useful since it does a whole-string match, whereas I want to only look for a substring, so I will stick with this implementation.

abayer:
:bee: Hurrah - been waiting for this.

kohsuke:
:bee:

svanoort:
🐝

abayer:
:bee:

roni-frantchi:
@jglick Many thanks for this one.
Looking at [your Stack answer](http://stackoverflow.com/questions/32418416/how-to-get-svn-changelog-in-jenkins-workflow-plugin/32525026#32525026), as well as [your issue](https://issues.jenkins-ci.org/browse/JENKINS-30412), I still can't quite figure out how one should use your fix.
Can you provide an example?..

Thanks.

PR #9, opened 2016-09-09
Add duration info to pipeline steps (tree view) UI

svanoort:
Adds timing info to tree view to help figure out what is taking time in a pipeline.  Spinoff from Jenkins world talks. 

- [x] Gets basic timing info for nodes
- [x] For blocks, they use the end node to compute total duration (yay!)
- [x] Handle cases where no start time is present
- [x] Provide a clean string formatting for elapsed time
- [x] Show if time is in the block or just the node self (cue for users on hoo to interpret)
- [x] Bugfix: figure out why last node or pair of nodes in a block is sometimes not getting timing info?

Manually tested with different (often complex) nesting structures, and with parallels.  Bug above is only one seen.  If no TimingAction is present on one of the flownodes (possible if there's a failure with loading actions), then no timing is shown for that one, rather than invalid info.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable.java
@@ -5,10 +5,8 @@ import org.jenkinsci.plugins.workflow.actions.TimingAction;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.graph.BlockEndNode;
 import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
-import org.jenkinsci.plugins.workflow.graph.FlowGraphWalker;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.actions.NotExecutedNodeAction;
-import org.jenkinsci.plugins.workflow.graph.FlowStartNode;
 import org.jenkinsci.plugins.workflow.graphanalysis.DepthFirstScanner;
 import org.jenkinsci.plugins.workflow.visualization.table.FlowNodeViewColumn;
 import org.jenkinsci.plugins.workflow.visualization.table.FlowNodeViewColumnDescriptor;
@@ -71,11 +69,11 @@ public class FlowGraphTable {
      */
     private Map<FlowNode, Row> createAllRows() {
         heads = execution.getCurrentHeads();
-        DepthFirstScanner scanner = new DepthFirstScanner();
+        final DepthFirstScanner scanner = new DepthFirstScanner();
         scanner.setup(heads);
 
         // nodes that we've visited
-        Map<FlowNode,Row> rows = new LinkedHashMap<FlowNode, Row>();
+        final Map<FlowNode,Row> rows = new LinkedHashMap<FlowNode, Row>();
 
         for (FlowNode n : scanner) {
             Row row = new Row(n);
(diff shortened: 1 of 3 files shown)


jglick:
Sounds useful.

svanoort:
**Screenshot in action:** 

screen shot 2016-09-09 at 4 47 27 pm

@reviewbybees

jglick:
> Show if time is in the block or just the node self

What about a step that takes some significant time in itself *and* in one or more blocks?

jglick:
Did you test with steps which run multiple blocks, such as {{waitUntil}} or {{retry}}?

jglick:
```
Nullcheck of newRow at line 217 of value previously dereferenced in org.jenkinsci.plugins.workflow.support.visualization.table.FlowGraphTable.order(FlowGraphTable$Row) [org.jenkinsci.plugins.workflow.support.visualization.table.FlowGraphTable, org.jenkinsci.plugins.workflow.support.visualization.table.FlowGraphTable] At FlowGraphTable.java:[line 217]Redundant null check at FlowGraphTable.java:[line 222] RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE
```

svanoort:
> What about a step that takes some significant time in itself and in one or more blocks?

@jglick Could be a good future enhancement.  We should talk about use cases to see if it is worth displaying specially (and how).  Adds some significant complexity though, so would want to do it with good reason only. 

> multiple blocks

Why would that be an issue, if it only looks at BlockStartNode & BlockEndNode?

> FindBugs

Already fixed.

jglick:
> Why would that be an issue

I have no idea how your implementation works, I am just asking whether you did any tests of this possibly interesting case—if so, your screenshot does not show it.

jglick:
:bee:

svanoort:
> I am just asking whether you did any tests of this possibly interesting case—if so, your screenshot does not show it.

I have not, but would expect the implementation to handle them seamlessly, and it turns out it does!  *grin*

screen shot 2016-09-09 at 5 07 46 pm

src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable.java
```diff
@@ -67,13 +71,13 @@ public void build() {
      */
     private Map<FlowNode, Row> createAllRows() {
         heads = execution.getCurrentHeads();
-        FlowGraphWalker walker = new FlowGraphWalker();
-        walker.addHeads(heads);
+        DepthFirstScanner scanner = new DepthFirstScanner();
```
oleg-nenashev: use final, Luke

abayer:
works for me :bee: - sorry I just had @jglick do a release of `workflow-support` like 30 minutes ago. =)

src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable.java
```diff
@@ -243,7 +279,27 @@ public String getDisplayName() {
...
+            return this.startTimeMillis;
+        }
+
+        public long getDurationMillis() {
+            return this.durationMillis;
+        }
+
+        public String getDurationString() {
+            if (this.durationMillis == 0) {
+                return "";
```
oleg-nenashev: Hmm, maybe "N/A"?

svanoort: It's hidden anyway (not shown if 0), so kind of irrelevant

oleg-nenashev:
🐝 anyway

src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable.java
```diff
@@ -67,13 +69,13 @@ public void build() {
      */
     private Map<FlowNode, Row> createAllRows() {
         heads = execution.getCurrentHeads();
-        FlowGraphWalker walker = new FlowGraphWalker();
-        walker.addHeads(heads);
+        final DepthFirstScanner scanner = new DepthFirstScanner();
```
jglick: Does not work; cf. #13.

PR #10, opened 2016-09-13
Moving some code relating to LogActionImpl into that source for better encapsulation

jglick:
@reviewbybees


abayer:
:bee: assuming tests pass, of course. =)

PR #11, opened 2016-09-16
Better display timing for steps/blocks with under 1 ms or no timing

svanoort:
Better way to handle timing display if TimingActions are missing for one of the nodes:

* If TimingAction missing, show "no timing"
* If Time <1 ms, show "<1 ms" (this is the lowest time resolution allowed here)

@reviewbybees


kzantow:
:bee:

abayer:
:bee:ssssss

PR #14, opened 2016-09-23
Use depthfirstscanner again

svanoort:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/16 now that the somewhat eccentric iteration order in FlowGraphWalker is maintained in DepthFirstScanner.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,7 +74,7 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.4-SNAPSHOT</version> <!-- Snapshot until enforced ordering in DepthFirstScanner is released -->
+            <version>2.4</version>
         </dependency>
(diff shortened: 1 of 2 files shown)


svanoort:
@reviewbybees  especially @jglick

pom.xml
```diff
@@ -74,7 +74,7 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.2</version>
+            <version>2.4-SNAPSHOT</version> <!-- Snapshot until enforced ordering in DepthFirstScanner is released -->
```
jglick: FWIW I find it helpful to link to the actual upstream PR—simplifies bookkeeping when there are numerous PRs in flight.

svanoort: good point -- I'm about to toggle it now though

PR #15, opened 2016-09-23
[JEP-210] Log handling rewrite

jglick:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/17. Subsumes #75.

[JEP-210](https://github.com/jenkinsci/jep/blob/master/jep/210/README.adoc)

- [X] basics
- [x] fix rendering of console output from _Pipeline Steps_
- [ ] ~way to efficiently get all output from a branch~
- [x] retain compatibility if only this plugin is updated and not `workflow-job`
- [x] optimize `AnnotatedLogAction.getLogText`
- [x] optimize `AnnotatedLogAction.strip`


jglick:
Not going to bother deploying upstream snapshots at this stage, so CI failures are expected.

src/main/java/org/jenkinsci/plugins/workflow/support/actions/AnnotatedLogAction.java
```diff
@@ -0,0 +1,327 @@
...
+        return new DecoratedTaskListener(raw, null, prefix(id));
+    }
+
+    /**
+     * Unlike {@link FlowNode#isRunning}, handles {@link BlockStartNode}s.
+     */
+    private static boolean isRunning(FlowNode node) {
+        if (node instanceof BlockStartNode) {
+            for (FlowNode head : node.getExecution().getCurrentHeads()) {
+                if (new LinearBlockHoppingScanner().findFirstMatch(head, Predicates.equalTo(node)) != null) {
```
jglick: Replace with `isActive` from https://github.com/jenkinsci/workflow-api-plugin/pull/45. Looks like this is not the first time I thought to replace the `DepthFirstScanner` with multiple calls to `LinearBlockHoppingScanner`!

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LessAbstractTaskListener.java
```diff
@@ -0,0 +1,107 @@
...
+import java.io.PrintStream;
+import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+/**
+ * Replacement for {@link StreamBuildListener} that delegates all methods to {@link #getLogger} rather than a private field.
+ * This allows it to be sanely subclassed, for example to delegate to another listener, or to use a serial form that does not involve {@link RemoteOutputStream}.
+ */
+@Beta // would be best done in AbstractTaskListener
```
jglick: https://github.com/jenkinsci/jenkins/pull/3122

src/main/java/org/jenkinsci/plugins/workflow/support/actions/AnnotatedLogAction.java
```diff
@@ -0,0 +1,309 @@
...
+    @Restricted(DoNotUse.class) // Jelly
+    public void writeLogTo(long offset, XMLOutput out) throws IOException {
+        getLogText().writeHtmlTo(offset, out.asWriter());
+    }
+
+    /**
+     * Creates a sink to print output from a step.
+     * Will use {@link LogActionImpl} if necessary.
+     * @param node a node which wishes to print output
+     * @param filter an optional log filter (must be {@link Serializable})
```
jglick: As in https://github.com/jenkinsci/workflow-job-plugin/pull/27 and https://github.com/jenkinsci/workflow-api-plugin/pull/17 and https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/62#issuecomment-372345921 we should probably introduce into `remoting` something like

```java
public interface Remotable extends Serializable {
    default Channel getChannelForSerialization() throws NotSerializableException {/*…*/}
}
public interface SerializableOnlyOverRemoting extends Remotable {
    // empty
}
```

svanoort:
@jglick Does this fix the issue we discussed *months* ago where it requires reading the entire logfile to retrieve the fragment for a single step?

jglick:
> reading the entire logfile to retrieve the fragment for a single step

That has since been refactored into https://github.com/jenkinsci/workflow-api-plugin/pull/17 where it now lives in `StreamLogStorage.stepLog`. Since streaming a single file is relatively cheap for I/O, and the filtering is simple, that is just left as is to keep things simple. Of course if you are using, e.g., CloudWatch Logs there is a completely different implementation (in that case using server-side filtering).

Rendering speed is anyway not a priority for JEP-210, compared to optimizing writing during the build.

src/main/java/org/jenkinsci/plugins/workflow/support/actions/AnnotatedLogAction.java
```diff
@@ -0,0 +1,159 @@
...
+import org.jenkinsci.plugins.workflow.log.LogStorage;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.DoNotUse;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+/**
+ * A marker for a node which had some log text.
+ */
+@Restricted(NoExternalUse.class) // for use from DefaultStepContext only
+public class AnnotatedLogAction extends LogAction implements FlowNodeAction, PersistentAction {
```
jglick: This should be renamed.

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogStorageAction.java
```diff
@@ -0,0 +1,159 @@
...
+        FlowExecutionOwner owner = node.getExecution().getOwner();
+        if (LogActionImpl.isOld(owner) || node.getAction(LogActionImpl.class) != null) {
+            return LogActionImpl.stream(node, filter);
+        } else {
+            if (node.getAction(LogStorageAction.class) == null) {
+                node.addAction(new LogStorageAction(node));
+            }
+            TaskListener listener = LogStorage.of(owner).nodeListener(node);
+            if (filter != null) {
+                if (filter instanceof Serializable) {
```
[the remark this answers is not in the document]
jglick: They _must_ be `Serializable` in order to work with Pipeline at all, since otherwise the attempt to save context to `program.dat` would explode. This also applies to `SimpleBuildWrapper.createLoggerDecorator`.

This is just a best effort to deal with unserializable filters. An implementation of JENKINS-30777 would add global filters to the mix, though again this could simply exclude non`Serializable` ones from the start. `Serializable` is a marker interface, which is an antipattern in Java (IMHO), because it is impossible for an implementation such as `MergedFilter` cannot conditionally indicate serializability depending on its delegates. In this case it would not be worth much, since again _all_ Pipeline filters are expected to be `Serializable`.

Really what we want to know is not just whether the filter is `Serializable`, which merely means that its instance fields can be saved and restored, but whether it is _remotable_—whether its `decorateLogger` could meaningfully run from an agent JVM. We have no API for that currently. There is the marker interface from @oleg-nenashev called `SerializableOnlyOverRemoting`, but implementing that implies that the class _cannot_ be serialized to `program.dat`, which is not true—we want to assert that an object can be _both_ serialized normally and remoted. That would be a new Remoting and thus core API which plugins providing filters would need to update to. Probably overkill for something which can anyway be tested just by using new `workflow-*` test deps and running something on `JenkinsRule.createSlave`.

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogStorageAction.java
```diff
@@ -0,0 +1,159 @@
...
+     * Will use {@link LogActionImpl} if necessary.
+     * @param node a node which wishes to print output
+     * @param filter an optional log filter (must be {@link Serializable})
+     * @return a stream
+     */
+    @SuppressWarnings("deprecation") // LogActionImpl here for backward compatibility
+    public static @Nonnull TaskListener listenerFor(@Nonnull FlowNode node, @CheckForNull ConsoleLogFilter filter) throws IOException, InterruptedException {
+        FlowExecutionOwner owner = node.getExecution().getOwner();
+        if (LogActionImpl.isOld(owner) || node.getAction(LogActionImpl.class) != null) {
+            return LogActionImpl.stream(node, filter);
```
jglick: Probably the `!FlowNode.active` closing logic from `LogActionImpl` should be moved here—for some implementations it may be significant (though not for `FileLogStorage`). Also rather than `PrintStream.close` it should check for `TaskListener & AutoCloseable`. And the expected close behavior for both the overall and step logs should be documented in `LogStorage`.

jglick: (resolved)

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -97,6 +99,33 @@
...
+                throw new IOException("cannot start writing logs to a finished node " + node + " " + node.getDisplayFunctionName() + " in " + node.getExecution());
+            }
+            listener = LogStorageAction.listenerFor(node, get(ConsoleLogFilter.class));
+            LOGGER.log(Level.FINE, "opened log for {0}", node.getDisplayFunctionName());
+            if (listener instanceof AutoCloseable) {
+                node.getExecution().addListener(new GraphListener.Synchronous() {
+                    @Override public void onNewHead(FlowNode newNode) {
+                        if (!node.isActive()) {
+                            node.getExecution().removeListener(this);
+                            LOGGER.log(Level.FINE, "closing log for {0}", node.getDisplayFunctionName());
```
svanoort: For a good time, try to disentangle what all the xxFunctionName/xxDisplayName function for a FlowNode do and which to use when. 

And by a "good time" I mean "if you're a masochist."

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -62,17 +67,14 @@
         if (key == EnvVars.class) {
             Run<?,?> run = get(Run.class);
             EnvironmentAction a = run == null ? null : run.getAction(EnvironmentAction.class);
-            EnvVars customEnvironment = a != null ? a.getEnvironment() : run.getEnvironment(get(TaskListener.class));
+            EnvVars customEnvironment = a != null ? a.getEnvironment() : run.getEnvironment(getExecution().getOwner().getListener());
```
jglick: Amended in #62 FTR.

PR #16, opened 2016-10-05
Ensure that the SemaphoreStep always gets its TimingAction & test this

svanoort:
Solves [JENKINS-38737](https://issues.jenkins-ci.org/browse/JENKINS-38737). 

Possibly a bit hacky, but the step is only used for testing. 

@reviewbybees especially @jglick

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/SemaphoreTimingActionTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/SemaphoreTimingActionTest.java
@@ -1,0 +1,59 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2016 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/SemaphoreTimingActionTest.java
```diff
@@ -0,0 +1,59 @@
...
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
+
+public class SemaphoreTimingActionTest {
```
jglick: Better to call it simply `SemaphoreStepTest`.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -152,7 +155,14 @@ public static void waitForStart(@Nonnull String k, @CheckForNull Run<?,?> b) thr
 
         @Inject(optional=true) private SemaphoreStep step;
 
+        @StepContextParameter
+        private transient FlowNode node;
+
         @Override public boolean start() throws Exception {
+            // Because
```
jglick: Because what?

svanoort:
@jglick The issue seems to be fairly specific to Semaphore implementation in this case, ergo a fix here

jglick:
> The issue seems to be fairly specific to Semaphore implementation in this case

But what is the root cause?

svanoort:
@jglick In this case, it appeared to be interference between the Semaphore GraphListener and the one in WorkflowRun.

jglick:
Well that [seems to be unused](https://github.com/search?utf8=%E2%9C%93&q=user%3Ajenkinsci+SemaphoreListener&type=Code&ref=searchresults) so I am not sure how it could hurt anything, though it ought to be deleted. What brought you to this conclusion?

svanoort:
@jglick The fix was tested and did indeed resolve the issue.  However it's been weeks so I don't remember the investigation details in depth.   IMO the fix here was sufficient for a test-only class where I could have simply done a workaround and moved on.   But if you want to dig deeper, feel free to pick up the issue yourself as the resident expert: https://issues.jenkins-ci.org/browse/JENKINS-38737  -- it's unassigned now.

jglick:
OK.

PR #17, opened 2016-10-14
Set up PauseAction to be a PersistentAction for better performance

svanoort:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/21

@reviewbybees


svanoort:
Note: the build failure here is just a dependency issue, closing and reopening to trigger rebuild

PR #19, opened 2016-10-18
No messages printed from block-scoped steps after they invoked their bodies

jglick:
Superseded by #15 but this is a far more conservative and localized change.

Background: in [JENKINS-34637](https://issues.jenkins-ci.org/browse/JENKINS-34637) I would like `TimeoutStepExecution` to print something to the build log when its body times out. Without this patch, that is impossible—the log message is discarded.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.workflow.support.actions;
 
 import com.google.common.base.Charsets;
+import com.google.common.base.Predicates;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.console.AnnotatedLargeText;
 import hudson.console.ConsoleLogFilter;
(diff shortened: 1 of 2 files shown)


stephenc:
:bee: as I understand it... but my comprehension is not at its best right nie

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
```diff
@@ -119,6 +142,20 @@ public FlowNode getParent() {
...
 
     /**
+     * Unlike {@link FlowNode#isRunning}, handles {@link BlockStartNode}s.
+     * Note that {@code WorkflowRun.copyLogs} will still currently refuse to consider steps
+     * which fail to obtain a {@link TaskListener} prior to invoking their body
+     * (for example by calling {@link StepContext#get} on demand rather than by using {@link StepContextParameter}).
+     */
+    private static boolean isRunning(FlowNode node) {
+        if (node instanceof BlockStartNode) {
+            return new LinearBlockHoppingScanner().findFirstMatch(node.getExecution().getCurrentHeads(), Predicates.equalTo(node)) != null;
```
jglick: > you are potentially scanning the flow back to the beginning of time

Right, because in

```groovy
node { // #1
  for (int i = 0; i < 1000; i++) {sh 'echo' /* #2 */}
}
```

the current head would include point 2 and we need to scan back to point 1 if this method is called. Now we could add caching etc. if there an API for it.

jglick:
Seems to have caused a regression in `EnvStepRunTest.parallel`; under investigation.

PR #21, opened 2016-10-19
Eagerly load actions on flownodes

svanoort:
SimpleXStreamFlowNodeStorage stores & loads FlowNode data as Tag objects, which include both the node and its actions, and are de/serialized together.  However, we normally throw away the actions when fetching just node data, and then wait for them to be requested to lazy-load the info. 

It turns out that almost every time we're going to need those actions (either for iteration via ForkScanner, for timing, etc).  The only case where we don't is loading the currentHeads for an execution -- and in that case, it's very cheap to just *not* throw away the actions.

This sets the actions for the flownode upon Tag loading rather than throwing them away, so we don't pay the deserialization & disk hit twice.


svanoort:
@reviewbybees

abayer:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -138,6 +141,14 @@ private Tag(@Nonnull FlowNode node, @Nonnull List<Action> actions) {
             this.actions = actions.isEmpty() ? null : actions.toArray(new Action[actions.size()]);
         }
 
+        private void storeActions() {  // We've already loaded the actions, may as well store them to the FlowNode
+            try {
+                FlowNode_setActions.invoke(this.node, actions());
+            } catch (InvocationTargetException|IllegalAccessException ex) {
+                throw new Error(ex);
```
rsandell: errors should be reserved for the vm.

svanoort: @rsandell Fair enough (I think it probably reflects something at that level if the class changes in such a way), but can just convert to a RunTimeException.

rsandell:
I don't see much loading going on here, only storage. 
:bug: for the throwing of `Error` though.

svanoort:
@rsandell Addressed your bug, also: the loading happens when someone tries to get the FlowNode actions (they were lazy-loaded before) -- and it in turn triggers a call to the loadActions method here. 

If we set the actions on the flownode using the (already-deserialized and loaded from disk) node file, we never end up calling that.

rsandell:
:bee:

PR #22, opened 2016-10-20
isRunning misbehaved on nodes inside parallel

jglick:
Regression from #19.

Without fix, would fail with

```
[test0 #1] java.lang.IllegalStateException: cannot start writing logs to a finished node org.jenkinsci.plugins.workflow.cps.nodes.StepStartNode[id=9]
[test0 #1] 	at org.jenkinsci.plugins.workflow.support.actions.LogActionImpl.<init>(LogActionImpl.java:110)
[test0 #1] 	at org.jenkinsci.plugins.workflow.support.actions.LogActionImpl.stream(LogActionImpl.java:81)
[test0 #1] 	at org.jenkinsci.plugins.workflow.support.DefaultStepContext.get(DefaultStepContext.java:73)
[test0 #1] 	at org.jenkinsci.plugins.workflow.support.actions.LogActionImplTest$ChattyStep$Execution.run(LogActionImplTest.java:100)
[test0 #1] 	at org.jenkinsci.plugins.workflow.support.actions.LogActionImplTest$ChattyStep$Execution.access$100(LogActionImplTest.java:75)
[test0 #1] 	at org.jenkinsci.plugins.workflow.support.actions.LogActionImplTest$ChattyStep$Execution$Callback.onSuccess(LogActionImplTest.java:108)
[test0 #1] 	at …
```

Regarding #19, if you disable the assertion and just use `FlowNode.isRunning`, the test case would fail looking for

    logging from LBBL with 0 commands to go

@reviewbybees esp. @svanoort

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,8 +28,8 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.11</version>
-        <relativePath />
+        <version>2.17</version>
+        <relativePath/>
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-support</artifactId>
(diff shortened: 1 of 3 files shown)


svanoort:
🐝  with full understanding.  Only one of the currentHeads is inspected with LinearScanners

jglick:
I have a patch under preparation to make that apparent.

jglick:
Will work on a local test case for this which exercises all the known use cases. In the meantime I need this released.

svanoort:
@jglick Could you use a lightly modified version of what I have in pipeline-graph-analysis:  https://github.com/jenkinsci/pipeline-graph-analysis-plugin/blob/master/src/test/java/org/jenkinsci/plugins/workflow/pipelinegraphanalysis/StatusAndTimingTest.java#L319  

I am testing very similar logic there.

jglick:
No, my needs here are rather different.

Test case almost ready, just doing some polishing.

svanoort:
Looks reasonable, waiting for CI though

jglick:
CI build passed, not sure why it did not update.

PR #24, opened 2016-11-02
Reflectively deduplicate the descriptorId and id fields for flownodes on deserialization

svanoort:
Should reduce memory footprint for Jenkins pipeline a bit and simplify garbage collection some. 

See [JENKINS-39456](https://issues.jenkins-ci.org/browse/JENKINS-39456)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
@@ -148,23 +148,71 @@ public class SimpleXStreamFlowNodeStorage extends FlowNodeStorage {
     private static final Field FlowNode$exec;
     private static final Field FlowNode$parents;
     private static final Field FlowNode$parentIds;
+    private static final Field FlowNode$id;
 
     static {
         XSTREAM.registerConverter(new Converter() {
             private final RobustReflectionConverter ref = new RobustReflectionConverter(XSTREAM.getMapper(), JVM.newReflectionProvider());
             // IdentityHashMap could leak memory. WeakHashMap compares by equals, which will fail with NPE in FlowNode.hashCode.
             private final Map<FlowNode,String> ids = CacheBuilder.newBuilder().weakKeys().<FlowNode,String>build().asMap();
+
+            private final Object NO_FIELD = new Object();
+
+            // Used to determine if we have a descriptorId field and if so, deduplicate them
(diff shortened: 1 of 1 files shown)


svanoort:
@reviewbybees especially @jglick

src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -148,23 +148,71 @@ private Tag(@Nonnull FlowNode node, @Nonnull List<Action> actions) {
...
+                        return;
+                    } else if (ob != null) {
+                        Field descriptorIdField = (Field)ob;
+                        Object fieldVal = descriptorIdField.get(f);
+                        if (fieldVal != null) {
+                            descriptorIdField.set(f, ((String)fieldVal).intern());
+                        }
+                    } else { // First time we saw this class
+                        try {
+                            Field field = c.getDeclaredField("descriptorId");
```
jglick: Way too much work. Just amend https://github.com/jenkinsci/workflow-cps-plugin/pull/82 to give `Step{Atom,Start}Node` a `readResolve`.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -148,23 +148,71 @@ private Tag(@Nonnull FlowNode node, @Nonnull List<Action> actions) {
...
                     FlowNode n = (FlowNode) ref.unmarshal(reader, context);
-                    ids.put(n, reader.getValue());
+                    String val = reader.getValue();
+                    if (val != null) {
+                        val = val.intern();
+                    }
+                    ids.put(n, val);
+
                     try {
+                        FlowNode$id.set(n, n.getId().intern());
```
jglick: I think this could be moved into `FlowNode.readResolve`.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
svanoort: That looks so much better, didn't know XStream invoked that -- if it works right (and it should) then I can bury this mess (waiting to see though).

svanoort:
Closing for https://github.com/jenkinsci/workflow-api-plugin/pull/23 (assuming it works right)

jglick:
Yeah that looks a lot simpler!

PR #27, opened 2017-01-02
Add getLog() to whitelist

ernetas:
It would be nice to get build logs without approving `method org.jenkinsci.plugins.workflow.support.steps.build.RunWrapper getRawBuild`, which is fairly permissive.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -162,6 +162,11 @@ public final class RunWrapper implements Serializable {
     }
 
     @Whitelisted
+    public List<String> getLog(int maxLines) throws IOException {
+        return build().getLog(maxLines);
+    }
+
+    @Whitelisted
     public @CheckForNull RunWrapper getPreviousBuild() throws AbortException {
         Run<?,?> previousBuild = build().getPreviousBuild();
         return previousBuild != null ? new RunWrapper(previousBuild, false) : null;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -162,6 +162,11 @@ public String getFullProjectName() throws AbortException {
     }
 
     @Whitelisted
+    public List<String> getLog(int maxLines) throws IOException {
+        return build().getLog(maxLines);
```
jglick: Possibly bad for performance if the argument is large. On the other hand sometimes you would want the full log. https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/2 might provide a more scalable solution, at least for the common case of scanning for messages in the current build’s log after an error (this method could be used for other purposes, such as after a `build` step).

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper/help.html
```diff
@@ -3,6 +3,7 @@
     <dt><code>result</code></dt><dd>typically <code>SUCCESS</code>, <code>UNSTABLE</code>, or <code>FAILURE</code> (<em>may</em> be null for an ongoing build)</dd>
     <dt><code>displayName</code></dt><dd>normally <code>#123</code> but sometimes set to, e.g., an SCM commit identifier</dd>
     <dt><code>description</code></dt><dd>additional information about the build</dd>
+    <dt><code>log</code></dt><dd>console output of the build</dd>
```
jglick: This is a method, not a property, so must be documented differently.

jglick:
cf. #51

dwnusbaum:
I am going to go ahead and close this. I think we should discuss this RFE on [JENKINS-46376](https://issues.jenkins-ci.org/browse/JENKINS-46376) to make sure we understand the use cases for a feature like this to see if we can do something that doesn't encourage users to process logs directly in their Pipelines.

PR #29, opened 2017-02-05
[JENKINS-32986] Introducing a general-purpose Timeout utility

jglick:
A component of [JENKINS-32986](https://issues.jenkins-ci.org/browse/JENKINS-32986).

@reviewbybees


jglick:
> a cleanup thread that triggers interrupt/hard-kill when it detects the pipeline execution has been explicitly terminated

Yes, also possibly useful (covered by other fixes in most cases), but independent of this.

PR #31, opened 2017-03-06
[FIXED JENKINS-42521] Add RunWrapper.getCurrentResult()

abayer:
[JENKINS-42521](https://issues.jenkins-ci.org/browse/JENKINS-42521)

getCurrentResult() will never be null - if build().getResult() is
null, it'll return Result.SUCCESS.toString(). In a perfect world, I'd
change RunWrapper.getResult() to this behavior, but that'd possibly
break existing logic out in the wild, so a new method it is.

cc @reviewbybees esp @jglick @rsandell @batmat @svanoort

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -128,6 +128,20 @@ public final class RunWrapper implements Serializable {
     }
 
     @Whitelisted
+    public boolean resultIsBetterOrEqualTo(String other) throws AbortException {
+        Result result = build().getResult() != null ? build().getResult() : Result.SUCCESS;
+        Result otherResult = Result.fromString(other);
+        return result.isBetterOrEqualTo(otherResult);
+    }
+
+    @Whitelisted
+    public boolean resultIsWorseOrEqualTo(String other) throws AbortException {
+        Result result = build().getResult() != null ? build().getResult() : Result.SUCCESS;
+        Result otherResult = Result.fromString(other);
+        return result.isWorseOrEqualTo(otherResult);
+    }
+
+    @Whitelisted
     public long getTimeInMillis() throws AbortException {
         return build().getTimeInMillis();
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -118,12 +118,6 @@ public final class RunWrapper implements Serializable {
     @Whitelisted
     public @CheckForNull String getResult() throws AbortException {
         Result result = build().getResult();
-        return result != null ? result.toString() : null;
-    }
-
-    @Whitelisted
-    public @Nonnull String getCurrentResult() throws AbortException {
-        Result result = build().getResult();
         return result != null ? result.toString() : Result.SUCCESS.toString();
     }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -129,14 +129,20 @@ public final class RunWrapper implements Serializable {
 
     @Whitelisted
     public boolean resultIsBetterOrEqualTo(String other) throws AbortException {
-        Result result = build().getResult() != null ? build().getResult() : Result.SUCCESS;
+        Result result = build().getResult();
+        if (result == null) {
+            result = Result.SUCCESS;
+        }
         Result otherResult = Result.fromString(other);
         return result.isBetterOrEqualTo(otherResult);
     }
 
     @Whitelisted
     public boolean resultIsWorseOrEqualTo(String other) throws AbortException {
-        Result result = build().getResult() != null ? build().getResult() : Result.SUCCESS;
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapperTest.java
```diff
@@ -166,6 +166,24 @@
         });
     }
 
+    @Issue("JENKINS-37366")
+    @Test public void getCurrentResult() {
+        r.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                MockFolder folder = r.j.createFolder("this-folder");
```
jglick: Why the folder?

abayer: That would just be copy-paste. Whoops. Fixing.

abayer:
@jglick D'oh, good point on the `getResult()` `null` - anything that's checking against `getResult()` being `null` now is almost certainly just using that as a `if (currentBuild.result == 'SUCCESS' || currentBuild.result == null)` type of situation. Will update that, and `help`.

svanoort:
@abayer I don't understand why we're defaulting to success here -- is there something we're missing?

abayer:
@svanoort So `run.getResult() == null` is effectively equivalent to "the run is successful so far", and for people trying to do things like send mail notifications with the run status in the subject, etc, they currently have to special case `currentBuild.result ?: 'SUCCESS'`, which is a common frustration.

orrc:
> No one should be relying on a null value

> anything that's checking against `getResult()` being `null` now is almost certainly just using that as a `if (currentBuild.result == 'SUCCESS' || currentBuild.result == null)` type of situation

I would challenge the "almost certainly" part of that statement.  Many people trying Pipeline will have found that almost always, `currentBuild.result` was `null` for their successful builds, and will have written their Pipeline scripts accordingly.

Certainly I've written a Pipeline shared library which started out only checking for `null`, since (virtually) no plugins ever explicitly set `SUCCESS`.  Though I did discover that one of the analysis-core plugins does indeed do that, so I fixed my code to check for `null || 'SUCCESS'`…

But a quick look through GitHub search shows there are pipelines which will break with this change:
* https://github.com/llvm-beanz/llvm-submodules/blob/31bf72e/Bots/FreeBSD-RA-x86_64-Submodule.groovy
* https://github.com/orientechnologies/orientdb-studio/blob/f8fddb8/Jenkinsfile
* https://github.com/reallyenglish/ansible-role-redis/blob/4e1c8fc/Jenkinsfile

Unfortunately, that last example which confuses `SUCCESSFUL` with `SUCCESS` was copy-pasted from the [Jenkins blog][0] (I'll file a PR to fix the affected posts), but otherwise all relevant content on jenkins.io that I could find does at least check for `null || 'SUCCESS'`.

Anyway, this PR is a good and sensible change — I did find a bunch of pipelines on GitHub that set `currentBuild.result = 'SUCCESS'` up-front, or did the `?: 'SUCCESS'` trick.

But when this gets released, we will need to be very clear in communicating to users that upgrading this plugin may break (or provide an opportunity to simplify!) existing pipelines.

[0]:https://jenkins.io/blog/2017/02/15/declarative-notifications/#moving-the-code-to-the-library

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -118,7 +118,7 @@ public int getNumber() throws AbortException {
     @Whitelisted
     public @CheckForNull String getResult() throws AbortException {
```
orrc: Annotation no longer applies.

daniel-beck:
Isn't this a good opportunity for terminology cleanup? The Jenkins UI generally uses "Success" to mean "Stable or Unstable".

jglick:
> a quick look through GitHub search shows there are pipelines which will break with this change

Hmm, those scripts are certainly wrong, but perhaps we need to go back to deprecating `result` and introducing a new property with the non-null behavior.

abayer:
Latest commit replaces the help text for `currentResult` and adds `resultIsBetterOrEqualTo(String)` and `resultIsWorseOrEqualTo(String)` methods to `RunWrapper` for [JENKINS-41275](https://issues.jenkins-ci.org/browse/JENKINS-41275)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -122,6 +122,26 @@ public int getNumber() throws AbortException {
...
 
     @Whitelisted
+    public @Nonnull String getCurrentResult() throws AbortException {
+        Result result = build().getResult();
+        return result != null ? result.toString() : Result.SUCCESS.toString();
+    }
+
+    @Whitelisted
+    public boolean resultIsBetterOrEqualTo(String other) throws AbortException {
+        Result result = build().getResult() != null ? build().getResult() : Result.SUCCESS;
```
jglick: This would be a race condition were it not for the fact that the `result` can go from null to non-null but not vice-versa.

abayer: That said, gonna still switch it to be less goofy.

rsandell:
@abayer so maybe we should now introduce a series of post conditions like `unstableOrWorse`, `unstableOrBetter` etc? :)

PR #33, opened 2017-03-21
[FIXED JENKINS-42952] Make currentBuild.duration meaningful

abayer:
[JENKINS-42952](https://issues.jenkins-ci.org/browse/JENKINS-42952)

...and add currentBuild.durationString.

cc @reviewbybees esp @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -133,7 +133,12 @@ public final class RunWrapper implements Serializable {
 
     @Whitelisted
     public long getDuration() throws AbortException {
-	 return build().getDuration();
+	 return System.currentTimeMillis() - build().getStartTimeInMillis();
+    }
+
+    @Whitelisted
+    public String getDurationString() throws AbortException {
+        return build().getDurationString();
     }
 
     @Whitelisted
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -133,7 +133,7 @@ public final class RunWrapper implements Serializable {
 
     @Whitelisted
     public long getDuration() throws AbortException {
-	 return System.currentTimeMillis() - build().getStartTimeInMillis();
+	 return build().getDuration() != 0 ? build().getDuration() : System.currentTimeMillis() - build().getStartTimeInMillis();
     }
 
     @Whitelisted
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapperTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapperTest.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.support.steps.build;
 
+import hudson.model.Messages;
 import hudson.model.Result;
 import java.util.regex.Pattern;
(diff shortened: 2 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -133,7 +133,12 @@ public long getStartTimeInMillis() throws AbortException {
 
     @Whitelisted
     public long getDuration() throws AbortException {
-	 return build().getDuration();
+	 return System.currentTimeMillis() - build().getStartTimeInMillis();
```
jglick: This will be wrong for a build other than `currentBuild`.

abayer: Updated.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapperTest.java
```diff
@@ -166,6 +167,21 @@
...
+    @Test public void duration() {
+        r.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = r.j.createProject(WorkflowJob.class, "this-job");
+                p.setDefinition(new CpsFlowDefinition(
+                        "echo \"currentBuild.duration='${currentBuild.duration}'\"\n" +
+                                "echo \"currentBuild.durationString='${currentBuild.durationString}'\"\n", true));
+                WorkflowRun b = r.j.assertBuildStatusSuccess(p.scheduleBuild2(0).get());
+                r.j.assertLogNotContains("currentBuild.duration='0'", b);
+                r.j.assertLogNotContains("currentBuild.durationString='" + Messages.Run_NotStartedYet() + "'", b);
```
jglick: Should this not be `Run_InProgressDuration`?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapperTest.java
```diff
@@ -166,6 +167,21 @@
...
 
+    @Issue("JENKINS-42952")
+    @Test public void duration() {
+        r.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                WorkflowJob p = r.j.createProject(WorkflowJob.class, "this-job");
+                p.setDefinition(new CpsFlowDefinition(
+                        "echo \"currentBuild.duration='${currentBuild.duration}'\"\n" +
+                                "echo \"currentBuild.durationString='${currentBuild.durationString}'\"\n", true));
+                WorkflowRun b = r.j.assertBuildStatusSuccess(p.scheduleBuild2(0).get());
```
jglick: `r.j.buildAndAssertSuccess(p)`

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapperTest.java
```diff
@@ -166,6 +167,21 @@
...
+                r.j.assertLogNotContains("currentBuild.duration='0'", b);
```
jglick: What if it in fact took a few milliseconds to get that far? Would the test not fail?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapperTest.java
abayer: re: this and the above - I think you're reading it wrong - it's `assertLogNotContains`, not `assertLogContains`.

jglick: Ah.

PR #34, opened 2017-03-21
[JENKINS-40934] Avoid redundant scans of parallels

svanoort:
[JENKINS-40934](https://issues.jenkins-ci.org/browse/JENKINS-40934)

Pretty straightforward fast fix here -- we could avoid a lot of this nonsense if our flows kept an easy trace of enclosing blocks in the context, but that's a much bigger change to make.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
@@ -152,13 +152,16 @@ public class LogActionImpl extends LogAction implements FlowNodeAction {
      * (for example by calling {@link StepContext#get} on demand rather than by using {@link StepContextParameter}).
      */
     private static boolean isRunning(FlowNode node) {
+        if (node.isRunning()) {
+            return true;
+        }
         if (node instanceof BlockStartNode) {
             // Block start is considered running if currently executing nodes are part of the block
             List<FlowNode> headNodes = node.getExecution().getCurrentHeads();
             AbstractFlowScanner scanner = (headNodes.size() > 1) ? new DepthFirstScanner() : new LinearBlockHoppingScanner();
-            return (scanner.findFirstMatch(headNodes, Predicates.equalTo(node)) != null) ? true : false;
+            return scanner.findFirstMatch(headNodes, Predicates.equalTo(node)) != null;
         } else {
-            return node.isRunning();
+            return false;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
```diff
@@ -149,12 +153,10 @@ public FlowNode getParent() {
...
         if (node instanceof BlockStartNode) {
-            for (FlowNode head : node.getExecution().getCurrentHeads()) {
-                if (new LinearBlockHoppingScanner().findFirstMatch(head, Predicates.equalTo(node)) != null) {
-                    return true;
-                }
-            }
-            return false;
+            // Block start is considered running if currently executing nodes are part of the block
+            List<FlowNode> headNodes = node.getExecution().getCurrentHeads();
+            AbstractFlowScanner scanner = (headNodes.size() > 1) ? new DepthFirstScanner() : new LinearScanner();
```
jglick: If there is just one head, `LinearBlockHoppingScanner` would be more efficient than `LinearScanner`, right? So you do not need to examine unrelated previous blocks.

`DepthFirstScanner` is not going to be terribly efficient for a big graph, but I suppose it will at least be faster than the current code.

svanoort: Yes, it is still O(n)ish and not amazing, but still much better with many parallels.  The "right" way is to store a datastructure holding the block structure of the flow that can be efficiently queried for enclosing blocks.

We've talked about that before, but let's not hold up a solid fix just because we want to reach an even better solution that may be much further off.

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
```diff
@@ -149,12 +153,10 @@ public FlowNode getParent() {
...
-            for (FlowNode head : node.getExecution().getCurrentHeads()) {
-                if (new LinearBlockHoppingScanner().findFirstMatch(head, Predicates.equalTo(node)) != null) {
-                    return true;
-                }
-            }
-            return false;
+            // Block start is considered running if currently executing nodes are part of the block
+            List<FlowNode> headNodes = node.getExecution().getCurrentHeads();
+            AbstractFlowScanner scanner = (headNodes.size() > 1) ? new DepthFirstScanner() : new LinearBlockHoppingScanner();
+            return (scanner.findFirstMatch(headNodes, Predicates.equalTo(node)) != null) ? true : false;
```
jglick: More simply:

```java
return scanner.findFirstMatch(headNodes, Predicates.equalTo(node)) != null;
```

PR #35, opened 2017-03-24
[JENKINS-26137] Pass pickleFactories directly into the RiverWriter constructor

jglick:
@reviewbybees


jglick:
It will certainly conflict with #26 but I think not in any deep sense.

PR #36, opened 2017-06-15
[FIXED JENKINS-41272] Expose build causes from currentBuild

mpavlov:
[JENKINS-41272](https://issues.jenkins-ci.org/browse/JENKINS-41272)
cc @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -26,10 +26,12 @@ package org.jenkinsci.plugins.workflow.support.steps.build;
 
 import hudson.AbortException;
 import hudson.model.AbstractBuild;
+import hudson.model.Cause;
 import hudson.model.Result;
 import hudson.model.Run;
 import hudson.scm.ChangeLogSet;
 import hudson.security.ACL;
+import hudson.triggers.TimerTrigger;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -240,4 +242,21 @@ public String getId() throws AbortException {
         }
     }
 
+    @Whitelisted
+    public List<Cause> getCauses() throws AbortException {
```
jglick: Only useful insofar as the `Cause` implementations themselves have whitelisted getters.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -240,4 +242,21 @@ public String getId() throws AbortException {
...
+        return build().getCauses();
+    }
+
+    @Whitelisted
+    public List<String> getCauseDescriptions() throws AbortException {
```
jglick: Would rather not expose this as it encourages scripts to rely on inherently unreliable localized messages.

mpavlov:
@jglick I see the point in not enabling functionality that's easily misused.

Would it still be beneficial to solve the specific problem (JENKINS-41272) separately from the general `DescribableModel`-like POJO uninstantiation system?

For example, maybe something like this could greatly help the majority of Cause-based use cases?
```
    @Whitelisted
    public boolean getIsStartedByUser() {
        return build().getCause(Cause.UserIdCause.class) != null;
    }

    @Whitelisted
    public boolean getIsStartedByTimer() {
        return build().getCause(TimerTrigger.TimerTriggerCause.class) != null;
    }
```

gabloe:
> For example, maybe something like this could greatly help the majority of Cause-based use cases?
It would be nice if the cause was an upstream project to be able to get the upstream project name as well.

mpavlov:
@jglick I've removed the reliance on localized messages. Can you take another look?
@gabloe I've added `getUpstreamProjectName()`
cc: @dtschan

smokris:
It would be helpful to also have a `getStartedByUserId()` method, so we can customize the build for the user who initiated it.

I don't have write access to this PR, but I think it would be something like this:

```groovy
@Whitelisted
public String getStartedByUserId() throws Exception {
    Cause.UserIdCause cause = build().getCause(Cause.UserIdCause.class);
    return cause == null ? null : cause.getUserId();
}
```

mpavlov:
Based on the comments, the proposed change is not a good direction for a long-term solution. And it appears there is not enough interest to use it as an interim fix either. Closing.

PR #38, opened 2017-07-27
[JENKINS-38223] Using FlowNode.isActive to improve JENKINS-45553

jglick:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/45 & https://github.com/jenkinsci/structs-plugin/pull/27. See discussion in [JENKINS-45553](https://issues.jenkins-ci.org/browse/JENKINS-45553).

@reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(jenkinsVersions: [null, '2.32.2'])
+buildPlugin(jenkinsVersions: [null, '2.60.2'])
--- a/pom.xml
+++ b/pom.xml
@@ -77,7 +77,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.20-20170731.163505-3</version> <!-- TODO https://github.com/jenkinsci/workflow-api-plugin/pull/45 -->
+            <version>2.20-20170803.185040-4</version> <!-- TODO https://github.com/jenkinsci/workflow-api-plugin/pull/45 -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -78,12 +78,11 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
             <version>2.20-20170731.163505-3</version> <!-- TODO https://github.com/jenkinsci/workflow-api-plugin/pull/45 -->
-            <exclusions>
-                <exclusion>
-                    <groupId>org.jenkins-ci.plugins</groupId>
-                    <artifactId>scm-api</artifactId>
-                </exclusion>
-            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>scm-api</artifactId>
+            <version>2.0.8</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -125,12 +124,6 @@
             <artifactId>workflow-cps</artifactId>
             <version>2.33</version>
             <scope>test</scope>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.jenkins-ci.plugins</groupId>
-                    <artifactId>scm-api</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -116,7 +116,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.10-20170731.160029-1</version> <!-- TODO https://github.com/jenkinsci/structs-plugin/pull/27 -->
+            <version>1.10</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
+        <jenkins.version>2.7.3</jenkins.version>
         <no-test-jar>false</no-test-jar>
         <git-plugin.version>3.0.5</git-plugin.version>
         <workflow-scm-step-plugin.version>2.4</workflow-scm-step-plugin.version>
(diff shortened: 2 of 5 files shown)


jglick:
Test failures, need to investigate.

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
```diff
@@ -94,7 +85,7 @@
         LOGGER.log(Level.FINE, "opened log for {0}", node.getDisplayFunctionName());
         graphListener.set(new GraphListener.Synchronous() {
             @Override public void onNewHead(FlowNode newNode) {
-                if (!isRunning(node)) {
+                if (!node.isActive()) {
```
svanoort: Love it.

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
```diff
@@ -110,8 +101,8 @@
     private String charset;
 
     private LogActionImpl(FlowNode parent, Charset charset) {
-        if (!isRunning(parent)) {
-            throw new IllegalStateException("cannot start writing logs to a finished node " + parent);
+        if (!parent.isActive()) {
+            LOGGER.log(Level.WARNING, null, new IllegalStateException("cannot start writing logs to a finished node " + parent + " " + parent.getDisplayFunctionName() + " in " + parent.getExecution()));
```
svanoort: Safe to swallow the exception?

svanoort: Mixed feelings on this @jglick -- think it's better to throw the exception here because then we can't create an invalid LogActionImpl.  Plus AFAICT the exception wouldn't be otherwise swallowed, and we can always try/catch it at the single constructor invocation if needed.

jglick: The exception is not “swallowed”, this is just a decision to log a stack trace.

My reasoning for switching from throwing an exception to just logging a warning was that an earlier mistake in the `isActive` implementation showed that the build could be completely broken by such a bug, which seems too harsh given that there is no real harm done if some content happens to be written to a step log after the step is supposedly complete. It may not show up in the build log but that is all. We do want to fix such bugs, hence the stack trace for diagnosis.

svanoort: You're making an exception go away regardless, you'll forgive me if my wording isn't precise late at night.  In any case, the question I'd ask is: is there a legit case for instantiating when the parent isn't active?  If so, warn but no exception thrown.  

If not, then throw exception but use try/catch in the flow elsewhere so we don't break the build (positive impact: prevents other errors from breaking the build).

src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable.java
```diff
@@ -260,7 +260,7 @@ private Row(FlowNode node) {
             if (act != null) {
                 this.startTimeMillis = act.getStartTime();
                 this.hasStartTime = true;
-                if (node.isRunning()) {
+                if (node.isActive()) {
```
jglick: Improves UI display of flow graph table.

src/test/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImplTest.java
```diff
@@ -58,27 +59,30 @@
         r.assertLogContains("atom step in A with 1 commands to go", b);
         r.assertLogContains("logging from LBBL with 0 commands to go", b);
         r.assertLogContains("atom step in B with 1 commands to go", b);
-        /* TODO misuse of FlowNode.isRunning in WorkflowRun.copyLogs prevents this from appearing unless we access TaskListener from start(); should be fixed by JENKINS-38381:
         r.assertLogContains("logging from BL with 0 commands to go", b);
-        */
     }
-    public static class ChattyStep extends AbstractStepImpl {
+    public static class ChattyStep extends Step {
```
jglick: just deprecations

src/test/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImplTest.java
```diff
@@ -58,27 +59,30 @@
         r.assertLogContains("atom step in A with 1 commands to go", b);
         r.assertLogContains("logging from LBBL with 0 commands to go", b);
         r.assertLogContains("atom step in B with 1 commands to go", b);
-        /* TODO misuse of FlowNode.isRunning in WorkflowRun.copyLogs prevents this from appearing unless we access TaskListener from start(); should be fixed by JENKINS-38381:
```
jglick: seems to more or less work now (JENKINS-38381 would fix ordering)

pom.xml
```diff
@@ -62,7 +62,7 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>1.642.3</jenkins.version>
+        <jenkins.version>2.7.3</jenkins.version>
```
jglick: To test against `workflow-cps` 2.33. No longer strictly necessary since we are no longer relying on `GraphListener` as an extension point, but it is high time we moved to Jenkins 2 anyway right?

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
jglick: > is there a legit case for instantiating when the parent isn't active?

No.

> warn but no exception thrown

That is what we are doing here.

> throw exception but use try/catch in the flow elsewhere so we don't break the build

Perhaps `DefaultStepContext.get` could throw this as an `IOException`, which in principle `StepExecution`s should already be catching.

Jenkinsfile
```diff
@@ -1 +1 @@
-buildPlugin(jenkinsVersions: [null, '2.32.2'])
+buildPlugin(jenkinsVersions: [null, '2.60.2'])
```
jglick: Or maybe time for `2.73`?

jglick:
Awaiting review on late changes.

svanoort:
I don't *think* there's anything new to re-review, but since you're asking now:  here's a question that I just thought of (sorry for late question): will this correctly handle someone doing a println in groovy after a step?

jglick:
> I don't think there's anything new to re-review

There were nontrivial code changes since your last review AFAICT.

> will this correctly handle someone doing a println in groovy after a step?

As per [this code](https://github.com/jenkinsci/workflow-cps-plugin/blob/459fe615077ee563ebf20561f8a6559b002ecfbf/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsScript.java#L205-L208) that is actually a step, same as writing `echo` explicitly.

svanoort:
ah, well okay then

jglick:
@reviewbybees done

PR #39, opened 2017-07-31
[JENKINS-37324] Add an ArgumentsColumn to the FlowGraphTable display

jglick:
screenshot

See comment in [JENKINS-37324](https://issues.jenkins-ci.org/browse/JENKINS-37324). Makes sense to use the new system in the rudimentary built-in UI.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/ArgumentsColumn.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/ArgumentsColumn.java
@@ -1,0 +1,58 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2017 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/ArgumentsColumn.java
```diff
@@ -0,0 +1,58 @@
...
+import hudson.Extension;
+import org.jenkinsci.plugins.workflow.actions.ArgumentsAction;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.visualization.table.FlowNodeViewColumn;
+import org.jenkinsci.plugins.workflow.visualization.table.FlowNodeViewColumnDescriptor;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+@Restricted(NoExternalUse.class)
```
oleg-nenashev: I am not sure if it makes the descriptor restricted. Likely no

jglick: If not, I guess that would be considered a bug in the access modifier restrictions library.

jglick:
@reviewbybees done

svanoort:
Love it!

PR #42, opened 2017-08-25
Optimization caching use in SimpleXStreamFlowNodeStorage

svanoort:
- [x] Switch to higher-performance Caffeine cache for FlowNodes
- [ ] Benchmark

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
@@ -24,9 +24,9 @@
 
 package org.jenkinsci.plugins.workflow.support.storage;
 
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
 import com.thoughtworks.xstream.converters.Converter;
 import com.thoughtworks.xstream.converters.MarshallingContext;
 import com.thoughtworks.xstream.converters.UnmarshallingContext;
@@ -52,7 +52,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ExecutionException;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -64,11 +61,9 @@
 public class SimpleXStreamFlowNodeStorage extends FlowNodeStorage {
     private final File dir;
     private final FlowExecution exec;
-    private final LoadingCache<String,FlowNode> nodeCache = CacheBuilder.newBuilder().softValues().build(new CacheLoader<String,FlowNode>() {
-        @Override public FlowNode load(String key) throws Exception {
-            return SimpleXStreamFlowNodeStorage.this.load(key).node;
-        }
-    });
+    private final LoadingCache<String,FlowNode> nodeCache = Caffeine.newBuilder().softValues().build(
+            key -> SimpleXStreamFlowNodeStorage.this.load(key).node
```
jglick: BTW `SimpleXStreamFlowNodeStorage.this.` could probably be elided

src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -164,8 +159,6 @@ private void storeActions() {  // We've already loaded the actions, may as well
     static {
         XSTREAM.registerConverter(new Converter() {
             private final RobustReflectionConverter ref = new RobustReflectionConverter(XSTREAM.getMapper(), JVM.newReflectionProvider());
-            // IdentityHashMap could leak memory. WeakHashMap compares by equals, which will fail with NPE in FlowNode.hashCode.
-            private final Map<FlowNode,String> ids = CacheBuilder.newBuilder().weakKeys().<FlowNode,String>build().asMap();
```
jglick: Why are you deleting the trick done in https://github.com/jenkinsci/pipeline-plugin/pull/213?

src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -175,17 +168,14 @@ private void storeActions() {  // We've already loaded the actions, may as well
...
                         @SuppressWarnings("unchecked") List<FlowNode> parents = (List<FlowNode>) FlowNode$parents.get(n);
                         if (parents != null) {
                             @SuppressWarnings("unchecked") List<String> parentIds = (List<String>) FlowNode$parentIds.get(n);
                             assert parentIds == null;
                             parentIds = new ArrayList<String>(parents.size());
                             for (FlowNode parent : parents) {
-                                String id = ids.get(parent);
-                                assert id != null;
-                                parentIds.add(id);
+                                parentIds.add(parent.getId().intern());
```
svanoort: May not even need to intern it since IIRC we use readResolve to intern the ID

jglick: The problem here is that when we are reading an old build record, `parent.getId() == null`.

jglick:
Basically the creepy code is supporting build records written by Workflow prior to 1.13. If you want to drop support for reading those build records, it is probably OK but you need to make sure the build is dropped gracefully—no NPEs.

jglick:
Basically any builds created by “Workflow” ([changelog](https://github.com/jenkinsci/pipeline-plugin/blob/master/CHANGES.md#113-jan-18-2016)).

jglick:
So, revert https://github.com/jenkinsci/pipeline-plugin/pull/213/commits/fcf87113f955bf9623e411be20ed0fe23ee22f36 and figure out how to detect the format and produce some clear error, for example throwing an exception from `SimpleXStreamFlowNodeStorage.<init>` which will be caught by `WorkflowRun.onLoad`. Easiest is for `FlowNode.readResolve` to check that `parents != null`, meaning we have old storage.

svanoort:
Should be rendered largely needless by https://github.com/jenkinsci/workflow-support-plugin/pull/47 assuming the alternate FlowNodeStorage implementation becomes common.

PR #43, opened 2017-09-01
Allow FlowNodeStorage to support in-memory Flownodes with deferred writing

svanoort:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/49

Upstream of some major performance optimizations that reduce redundant serialization calls by waiting until a FlowNode has its basic set of actions before persisting the whole thing. 

Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/49

TODO:
- [ ] Additional test coverage of revised storage engine (needed here, to explicitly exercise new logic even if the test version of workflow-cps doesn't include these goodies)

Deployed snapshot as:
```
<dependency>
            <groupId>org.jenkins-ci.plugins.workflow</groupId>
            <artifactId>workflow-support</artifactId>
            <version>2.15-deferactionwrite-20170915.012921-2-</version>
        </dependency>
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-support</artifactId>
-    <version>2.15-SNAPSHOT</version>
+    <version>2.15-deferactionwrite-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Supporting APIs</name>
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Supporting+APIs+Plugin</url>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -66,18 +66,20 @@
         <no-test-jar>false</no-test-jar>
         <git-plugin.version>3.0.5</git-plugin.version>
         <workflow-scm-step-plugin.version>2.4</workflow-scm-step-plugin.version>
-        <workflow-step-api-plugin.version>2.10</workflow-step-api-plugin.version>
+        <workflow-step-api-plugin.version>2.13-deferactionwrite2-20170901.210414-1</workflow-step-api-plugin.version>
```
svanoort: Probably not required but I'm being a bit extra cautious to ensure these dependencies always are added together.  I don't want maven to be "creative."

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -61,7 +61,7 @@
         T value = doGet(key);
         if (key == EnvVars.class) {
             Run<?,?> run = get(Run.class);
-            EnvironmentAction a = run.getAction(EnvironmentAction.class);
+            EnvironmentAction a = run == null ? null : run.getAction(EnvironmentAction.class);
```
svanoort: findbugs fix

svanoort:
@jglick Do you have any thoughts on what an effective test for this would look like?

svanoort:
(More specifically the suite of things)

svanoort:
Closing with preference to #47

PR #47, opened 2017-10-04
Granular flownode persistence and new node storage engine [JENKINS-47172] [JENKINS-47173]

svanoort:
Add granular persistence control for flownodes, support deferred writing flownode actions, and add a whole new speedy flownode storage engine. 

Implements [JENKINS-47172](https://issues.jenkins-ci.org/browse/JENKINS-47172) and [JENKINS-47173](https://issues.jenkins-ci.org/browse/JENKINS-47173)

Deployed as: workflow-support-2.17-durability-v4-20171220.174314-1

# Downstreams

* https://github.com/jenkinsci/workflow-cps-plugin/pull/188
* https://github.com/jenkinsci/workflow-job-plugin/pull/75
* https://github.com/jenkinsci/workflow-multibranch-plugin/pull/68


src/main/java/org/jenkinsci/plugins/workflow/support/storage/LumpFlowNodeStorage.java
```diff
@@ -0,0 +1,271 @@
...
+    public void flush() throws IOException {
+        if (nodes != null && isModified) {
+            // TODO reuse a single buffer if we can, and consider using async FileChannel operations.
+            if (!dir.exists()) {
+                IOUtils.mkdirs(dir);
+            }
+            OutputStream os = new BufferedOutputStream(
+                    Files.newOutputStream(getStoreFile().toPath(), StandardOpenOption.WRITE, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)
+            );
+            XSTREAM.toXMLUTF8(nodes, os); // Hah, no atomic nonsense, just write and write and write!
```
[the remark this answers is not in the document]
svanoort: @jtnord Yes, it will error on load -- this is for cases where the user has explicitly chosen to prioritize performance over absolute integrity (for that build). 

In practice, my suspicion is that this will be a rather rare event -- it requires interruption during a fairly brief write operation that is performed only a few times per build.  Also there's downstream testcases to verify that we handle that case cleanly.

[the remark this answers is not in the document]
jtnord: ack.

svanoort:
@reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -61,7 +61,7 @@
         T value = doGet(key);
         if (key == EnvVars.class) {
             Run<?,?> run = get(Run.class);
-            EnvironmentAction a = run.getAction(EnvironmentAction.class);
+            EnvironmentAction a = run == null ? null : run.getAction(EnvironmentAction.class);
```
abayer: We can get a null `Run` here now? How does that happen? I’m a bit nervous due to how many places assume you can never actually get a null `Run` from a `StepContext`.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -0,0 +1,247 @@
...
+        this.setAvoidAtomicWrite(true);
+    }
+
+    /** Loads the nodes listing, lazily - so loading the {@link FlowExecution} doesn't trigger a more complex load. */
+    HashMap<String, Tag> getOrLoadNodes() throws IOException {
+        if (nodes == null) {
+            if (dir.exists()) {
+                File storeFile = getStoreFile();
+                if (storeFile.exists()) {
+                    HashMap<String, Tag> roughNodes = (HashMap<String, Tag>) (XSTREAM.fromXML(getStoreFile()));
```
abayer: Redundant parens?

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -0,0 +1,247 @@
...
+                    if (roughNodes == null) {
+                        throw new IOException("Unable to load nodes, invalid data");
+                    }
+                    for (Tag t : roughNodes.values()) {
+                        FlowNode fn = t.node;
+                        try {
+                            FlowNode$exec.set(fn, exec);
```
abayer: Probably my ignorance speaking, but what’s with the dollar sign?

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    /**
+     * Persists a single FlowNode to disk (if not already persisted).
+     * @param n Node to persist
+     * @throws IOException
+     */
+    @Override
+    public void flushNode(@Nonnull FlowNode n) throws IOException {
+        storeNode(n, false);
```
abayer: My gut instinct on identical methods like this and `storeNode(FlowNode)` above is to have one call the other. No big deal, I just find it prevents later confusion.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        public @Nonnull List<Action> actions() {
+            return actions != null ? Arrays.asList(actions) : Collections.<Action>emptyList();
+        }
+    }
+
+    public static final XStream2 XSTREAM = new XStream2();
+
+    private static final Field FlowNode$exec;
```
abayer: Ah, ignore my earlier dollar sign question. That said, I’d prefer if we didn’t use uncommon characters and formatting in variable names.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/FlowNodeStorage.java
```diff
@@ -28,26 +28,87 @@
...
- * from storage mechanism, but not sure if this should be exposed to users.
+ * A flow graph has a characteristic that it is additive.
  *
+ * Flow nodes may be stored in memory or directly persisted to disk at any given moment, but invoking {@link #flush()}
+ *  should always guarantee that everything currently in memory is written.
  * @author Kohsuke Kawaguchi
+ * @author Sam Van Oort
  */
 public abstract class FlowNodeStorage implements FlowActionStorage {
+    // Set up as "avoid" because an unset field will default to false.
```
abayer: Seems to contradict the actual value?

src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -70,49 +75,125 @@
...
+                if (maybeOutput != null) {
+                    return maybeOutput;
+                }
+            }
             return nodeCache.get(id);
         } catch (ExecutionException x) {
-            throw new IOException(x); // could unwrap if necessary
+            Throwable cause = x.getCause();
+            if (cause instanceof NoSuchFileException) {
+                // No file, no node
```
abayer: Should there be some sort of logging for this scenario?

src/main/java/org/jenkinsci/plugins/workflow/support/storage/FlowNodeStorage.java
```diff
@@ -28,26 +28,87 @@
...
+
+    /** Registers node in this storage, potentially persisting to disk.
+     *  {@link #flushNode(FlowNode)} will guarantee it is persisted.
+     */
+    public abstract void storeNode(@Nonnull FlowNode n) throws IOException;
+
+    /**
+     * Register the given node to the storage, potentially flushing to disk,
+     *  and optionally marking the node as deferring writes.
+     * <p> This should be invoked with delayWritingAction=true generally.
```
abayer: Seems like this is false more often.

src/test/java/org/jenkinsci/plugins/workflow/ScalabilityTest.java
```diff
@@ -1,70 +0,0 @@
...
-import org.jenkinsci.plugins.workflow.job.WorkflowJob;
-import org.jenkinsci.plugins.workflow.job.WorkflowRun;
-import static org.junit.Assert.*;
-import org.junit.Test;
-import org.junit.Rule;
-import org.junit.runners.model.Statement;
-import org.jvnet.hudson.test.Issue;
-import org.jvnet.hudson.test.RestartableJenkinsRule;
-
-public class ScalabilityTest {
```
abayer: Why is this test being removed?

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
svanoort: It shouldn't but it came up in tests once, so I bulletproofed it -- but that's likely some bad code that was now fixed.  Left it in because... better nullchecks, why not?

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
svanoort: Nah, I just like to enclose things I'm casting to help visually separate the thing being cast.

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
abayer: Okiedokie. Just making sure it wasn't an expected result of a change.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
abayer: Fair enough!

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
svanoort: Nomenclature from SimpleXStreamFlowNodeStorage for a reflective field access when needed.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/FlowNodeStorage.java
svanoort: @abayer Not sure what you mean, but clarifying the  comment, thanks

abayer: The impression I got from reading the comment was that you'd be setting the field as true here, since if the default value for an unset field is false, why bother mentioning it when you're explicitly setting it to false anyway? Also, since you're setting it to false, you're setting it up as "don't avoid". =)

src/test/java/org/jenkinsci/plugins/workflow/ScalabilityTest.java
svanoort: It's long, slow, testing for one specific bug that can't recur, and the class of issue it was intended to cover is best covered with something like Hydra -- found an issue the other which this does absolutely *nothing* to cover even though this kind of test is supposed to, because it's too far up in the dependency tree.

It turns out that this sort of test needs to be on pipeline as a whole, really.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
svanoort: done, good point!

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
svanoort: And yeah, I don't completely love the special characters but it's helpful here to signal visually "hold on a second, you need to look closer here"

src/test/java/org/jenkinsci/plugins/workflow/ScalabilityTest.java
abayer: Fair point. I just get an initial queasy reaction to tests being removed. =)

svanoort: Dude, that is a totally healthy reaction to have.  I should have explained it. :)

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -149,7 +149,7 @@ public void storeNode(@Nonnull FlowNode n, boolean delayWritingActions) throws I
 
     @Override
     public void storeNode(FlowNode n) throws IOException {
-        storeNode(n, false);
+        flushNode(n);
```
abayer: Tiny quibble(s) - given that we don't have `@Nonnull FlowNode n` here, it feels like the call should be the other way around. But then, the abstract of `storeNode(FlowNode)` does have `@Nonnull FlowNode n`, so I guess really the answer is to put the annotation here too. Either way it's nitpicking at best. =)

src/main/java/org/jenkinsci/plugins/workflow/support/PipelineIOUtils.java
```diff
@@ -0,0 +1,38 @@
...
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+
+/**
+ * Utilities to assist with IO and in some cases improve performance specifically for pipeline.
+ */
+public class PipelineIOUtils {
+    /**
+     * Convenience method to allow writing a file more or less directly by reusing
```
jglick: …what?

src/main/java/org/jenkinsci/plugins/workflow/support/PipelineIOUtils.java
```diff
@@ -0,0 +1,38 @@
...
+     * @param xstream xstream to use for output
+     * @param atomicWrite If true, do an atomic write, otherwise do a direct write to file.
+     * @throws IOException
+     */
+    public static void writeByXStream(@Nonnull Object toWrite, @Nonnull File location, @Nonnull XStream2 xstream, boolean atomicWrite) throws IOException {
+        if (atomicWrite) {
+            XmlFile file = new XmlFile(xstream, location);
+            file.write(toWrite);
+        } else {
+            OutputStream os = new BufferedOutputStream(
```
jglick: `try`-with-resources

src/main/java/org/jenkinsci/plugins/workflow/support/PipelineIOUtils.java
```diff
@@ -0,0 +1,38 @@
...
+                    Files.newOutputStream(location.toPath(), StandardOpenOption.WRITE, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)
```
jglick: wrap `InvalidPathException`

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
```diff
@@ -58,7 +59,7 @@
  *
  * @author Kohsuke Kawaguchi
  */
-public class LogActionImpl extends LogAction implements FlowNodeAction {
+public class LogActionImpl extends LogAction implements FlowNodeAction, PersistentAction {
```
jglick: Fine but could be filed separately.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -0,0 +1,247 @@
...
+import java.lang.reflect.Method;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+
+/**
+ * {@link FlowNodeStorage} implementation that stores all the {@link FlowNode}s together in one file for efficient bulk I/O
```
jglick: BTW summary sentences must end in a period or your Javadoc display will be garbled.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
jglick: DANGER WILL ROBINSON @kohsuke CODE AHEAD

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -0,0 +1,247 @@
...
+    private static final Field FlowNode$exec;
+    private static final Method FlowNode_setActions;
+
+    static {
+        try {
+            // Ugly, but we do not want public getters and setters for internal state on FlowNodes.
+            FlowNode$exec = FlowNode.class.getDeclaredField("exec");
+            FlowNode$exec.setAccessible(true);
+            FlowNode_setActions = FlowNode.class.getDeclaredMethod("setActions", List.class);
+            FlowNode_setActions.setAccessible(true);
```
jglick: Uh, you have a snapshot dep open on `workflow-api` right now, so why not fix this once and for all and just introduce the public accessors (with stern warnings about no outside use)?

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -0,0 +1,247 @@
...
+        } catch (NoSuchFieldException|NoSuchMethodException e) {
+            throw new Error(e);
```
jglick: `ExceptionInInitializerError` FWIW

src/test/java/org/jenkinsci/plugins/workflow/ScalabilityTest.java
jglick: I think I actually suggested deleting this now. Appropriate at a higher level, and not in JUnit format since it is not pass/fail. (Well…actually it _did_ once fail, due a long-since-fixed bug.)

src/test/java/org/jenkinsci/plugins/workflow/support/storage/MockFlowExecution.java
```diff
@@ -0,0 +1,108 @@
...
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Mock {@link org.jenkinsci.plugins.workflow.flow.FlowExecution} for testing storage
+ * @author Sam Van Oort
+ */
+ // TODO the
```
jglick: ?

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
jglick: It _is_ correct to check for null here. Most places calling `get(Run.class)` do not need to, because `Run` was already declared part of the required context. Of course there are no production implementations of `FlowExecutionOwner` that are not based on `Run`, so this would never have come up. I suppose you hit this due to use of mocks.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
svanoort: Better to use horrifying reflection than have someone shoot themselves in the foot with that one though.

src/main/java/org/jenkinsci/plugins/workflow/support/PipelineIOUtils.java
```diff
@@ -9,14 +9,15 @@
...
+import java.nio.file.InvalidPathException;
 import java.nio.file.StandardOpenOption;
 
 /**
  * Utilities to assist with IO and in some cases improve performance specifically for pipeline.
  */
 public class PipelineIOUtils {
     /**
-     * Convenience method to allow writing a file more or less directly by reusing
+     * Convenience method to transparently write data directly or atomicly using {@link hudson.util.AtomicFileWriter}.
```
jglick: sp.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -234,6 +229,13 @@ private void storeActions() {  // We've already loaded the actions, may as well
     private static final Method FlowNode_setActions;
 
     static {
+        // Aliases reduce the amount of data persisted to disk
+        XSTREAM.alias("Tag", Tag.class);
```
jglick: This will render older versions of the plugin unable to read builds created by newer versions. That is not inherently a problem, but deserves a [`compatibleSinceVersion` note in the update center](http://jenkinsci.github.io/maven-hpi-plugin/jar-mojo.html#compatibleSinceVersion).

svanoort: Yep, it'll get some sort of note -- the Changlog is also a good place. 

Worth it because this reduces the amount of FlowNode data written quite significantly (30% or more).  For performance, this isn't as important as cutting the number of IO calls made, but it also reduces size-on-disk for the flowgraph.

PR #48, opened 2017-12-12
Make Timeout repeatedly interrupt a stuck thread, and note that this is happening

jglick:
Saw a support bundle that showed a thread inside

```
"jenkins.util.Timer [#3] / waiting for JNLP4-connect connection from …" …
   java.lang.Thread.State: TIMED_WAITING (on object monitor)
    at java.lang.Object.wait(Native Method)
    at hudson.remoting.Request.call(Request.java:165)
    - locked <0x…> (a hudson.remoting.UserRequest)
    at hudson.remoting.Channel.call(Channel.java:903)
    at hudson.FilePath.act(FilePath.java:987)
    at hudson.FilePath.act(FilePath.java:976)
    at hudson.FilePath.deleteRecursive(FilePath.java:1178)
    at org.jenkinsci.plugins.durabletask.FileMonitoringTask$FileMonitoringController.cleanup(FileMonitoringTask.java:187)
    at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.check(DurableTaskStep.java:326)
    at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.run(DurableTaskStep.java:279)
    at …
```

From that it was unclear whether this one call to `FilePath.deleteRecursive` was in fact taking more than the 10s allotted, or if there were lots of repeated calls to `run`/`check`, or what. This patch

* Ensures that even if some code swallows `InterruptedException`, a new signal will be sent.
* Makes sure something appears in the system log when a thread is blowing past its timeout.

I tried to also set the thread name to indicate how stalled it is, to allow this information to appear in thread dumps, but could not get it to work—whether due to a JVM bug or some oversight on my part, I am not sure.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.33</version>
+        <version>3.2</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -36,7 +36,7 @@
     <version>2.17-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Supporting APIs</name>
-    <url>https://wiki.jenkins-ci.org/display/JENKINS/Pipeline+Supporting+APIs+Plugin</url>
+    <url>https://wiki.jenkins.io/display/JENKINS/Pipeline+Supporting+APIs+Plugin</url>
     <licenses>
         <license>
             <name>MIT License</name>
@@ -63,6 +63,7 @@
     </pluginRepositories>
     <properties>
         <jenkins.version>2.60.2</jenkins.version>
+        <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <git-plugin.version>3.0.5</git-plugin.version>
         <workflow-scm-step-plugin.version>2.4</workflow-scm-step-plugin.version>
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Timeout.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Timeout.java
@@ -25,8 +25,8 @@
 package org.jenkinsci.plugins.workflow.support.concurrent;
 
 import hudson.FilePath;
+import hudson.Util;
 import hudson.remoting.VirtualChannel;
-import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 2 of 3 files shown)


oleg-nenashev:
> From that it was unclear whether this one call to FilePath.deleteRecursive was in fact taking more than the 10s allotted, or if there were lots of repeated calls to run/check, or wha

Object hashes were same across 9 threaddumps with a 5-sec interval, so most likely this is the first case

jglick:
> Object hashes were same

Which object hashes? `UserRequest`? If so, that suggests that `Channel.call` is ignoring thread interruptions. I do not [see that happening](https://github.com/jenkinsci/remoting/blob/cdec8f8aa6d52322bf6948f40c4cf82ddcf30dc1/src/main/java/hudson/remoting/Request.java#L154-L206) but maybe I am missing something.

Anyway, @reviewbybees done for now.

jglick:
By the way before implementing this (so using the original `Timeout` behavior) I did some crude testing: ran `ExecutorStepTest.buildShellScriptAcrossRestart`, which launches a JNLP agent and runs a `sh` on it, but modified to `sleep` longer to give me time to intervene; while the process was running, `STOP`ped the agent JVM so it would cease to respond; and after getting the expected

    Cannot contact dumbo: java.lang.InterruptedException

message, waited. The timeout was enforced to the extent that individual calls to `check` in fact ran for just 10s. The build continued to wait for a response (`check`ing repeatedly), as is currently designed—Pipeline does not treat a hung or even crashed agent JVM as fatal, since the same JVM or a replacement could come back and resume status updates about the external process. (I am not sure if @svanoort has plans to change this behavior or offer alternative durability strategies.) I then interrupted the build through the UI and it correctly `ABORTED` after printing

    Could not connect to dumbo to send interrupt signal to process

with no further human intervention.

Could probably develop this into an automated test in `workflow-durable-task-step` with a little work, though I am more interested in finding scenarios where recovery does _not_ happen as expected. In other words the repeated interruption portion of this patch is a speculative fix, not something I was able to demonstrate is necessary; the enhanced logging could in the future be used to rule out possible explanations for hangs.

abayer:
Fwiw, I think this is a problem that shows up on ci.jenkins.io pretty often, so a fix is nice to see.

jglick:
> it's hard to tell what the impact will be until we see it in the wild

Correct.

> I'm slightly nervous about using Timer threads more

Not sure what you mean. `Timeout` was already using `Timer` threads to deliver interrupts. This patch could cause _more_ tasks to be scheduled (if something else is indeed very wrong), but each one should run essentially instantaneously, and nothing waits for it.

> think hard about simplifying the pipeline threading model

If you have any suggestions let me know, but note that https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/21 would eliminate most of the `check`s discussed here (I left in a 5m ping to allow recovery from certain corner cases).

> this is a problem that shows up on ci.jenkins.io pretty often

If you have the ability to track down particular instances and investigate, that would be important work. I for one have no permissions on the server.

Anyway, @reviewbybees done for now.

jglick:
> taking more than the 10s allotted

Turns it out _was_ taking longer than 10s—but no interruption was being delivered, because the other nine `Timer` threads were also occupied! Oops.

jglick:
@reviewbybees done FWIW

jglick:
BTW the actual case of the agent-side hang was believed to be [JENKINS-39179](https://issues.jenkins-ci.org/browse/JENKINS-39179) in this case.

PR #51, opened 2018-01-12
[FIXED JENKINS-46376] Add RunWrapper#getLog()

abayer:
[JENKINS-46376](https://issues.jenkins-ci.org/browse/JENKINS-46376)

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -35,6 +35,7 @@ import hudson.security.ACL;
 import hudson.security.ACLContext;
 import java.io.IOException;
 import java.io.Serializable;
+import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -165,6 +166,17 @@ public String getDurationString() throws AbortException {
...
     }
 
+    @Whitelisted
+    public String getLog() throws AbortException {
+        StringWriter writer = new StringWriter();
+        try {
+            build().getLogText().writeLogTo(0, writer);
+            return writer.toString();
+        } catch (IOException e) {
+            throw new AbortException("Could not read log for build " + externalizableId + ": " + e);
```
jglick: This discards the original stack trace, and is really unnecessary. Just let the method throw up `IOException`.

abayer:
Yeah, I'm really not sure about it myself. I'll let it sit and think on it.

dwnusbaum:
I am going to go ahead and close this. I think we should discuss this feature on [JENKINS-46376](https://issues.jenkins-ci.org/browse/JENKINS-46376) to make sure we understand the use cases for a feature like this to see if we can do something that doesn't encourage users to process logs directly in their Pipelines.

PR #52, opened 2018-01-18
Whitelist the isKeepLog and setKeepLog properties (JENKINS-49014)

tknerr:
Exposes the `currentBuild.keepLog` getter / setter so that we can "keep this build forever" from within a pipeline script without accessing `currentBuild.rawBuild` (which would be blocked by script security)

See https://issues.jenkins-ci.org/browse/JENKINS-49014


tknerr:
@abayer @jglick is there any documentation that needs to be updated along with this?

tknerr:
Thanks! :)

tknerr:
@abayer curious: do you plan to ship this by an upcoming release anytime soon?

gschwienbacher:
@abayer: same here: we're interested too in what the plans for shipment in the upcoming releases are. thx.

PR #53, opened 2018-01-19
Ensure that regardless of where Timeout is first used, its executor has the correct classloader on the Thread

svanoort:
Fixes a minor but obnoxious bug where the Timeout utility is first invoked from inside Pipeline CpsThread and gets the wrong classloader... which means a leaky GroovyClassloader.

@reviewbybees


svanoort:
@jglick It most surely does resolve it

PR #56, opened 2018-02-09
[JEP-206] Always use UTF-8 for per-step log files

jglick:
[JEP-206](https://github.com/jenkinsci/jep/blob/master/jep/206/README.adoc)

Not exactly pulled out of #15, but in the same vein and complementary. One component of the solution only; as noted in https://github.com/jenkinsci/workflow-job-plugin/pull/89, users will need to update multiple plugins to resolve issues.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
@@ -38,6 +38,7 @@ import java.io.OutputStream;
 import java.io.PrintStream;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
```diff
@@ -162,8 +161,11 @@ public void onLoad(FlowNode parent) {
     }
 
     private Charset getCharset() {
-        if(charset==null)   return Charset.defaultCharset();    // just being defensive
-        return Charset.forName(charset);
+        if (charset == null) { // new style
+            return StandardCharsets.UTF_8;
```
svanoort: Is this actually safe to do?  Wouldn't some legacy builds potentially be in other Charsets?  

Seems like we should default the same way we did -- but always initialize with UTF-8, or am I missing something here?  (Granted I'm pretty out of it at the moment, so that's possible.)

Some of this might be why windows breaks.

jglick: Yes historical builds _would_ typically be in some other charset—which is why their `charset` field would be set, and we would honor that setting. It is only new builds which omit the field and thus use UTF-8 regardless of the master’s system default encoding.

jglick:
The Windows test failure is a bit misleading. It happened to pass before because the character being expected, `»`, is included in the ISO-8859-1 system encoding used by the CI build machine. With the complete fix, _all_ characters may be used in log files, regardless of environment.

jglick:
> the ISO-8859-1 system encoding used by the CI build machine

Well, more likely CP-1252 actually, but same applies.

jglick:
> the full stack

To be precise: https://github.com/jenkinsci/workflow-job-plugin/pull/89 (and this, its dependency), plus https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/64 (and https://github.com/jenkinsci/durable-task-plugin/pull/61, _its_ dependency). As a user, since the plugin manager deals with dependencies transparently, it means that you should update both **Pipeline: Job** and **Pipeline: Nodes and Processes** if you deal with funny characters. The fixes will apply to newly started builds (nothing should change for historical builds).

PR #57, opened 2018-03-19
[JENKINS-45693] Apply ConsoleLogFilter

mwinter69:
So far pipeline script ignore globally defined ConsoleLogFilters
probably because there hadn't been for long time. But now we have
(logstash, mask passwords) and they should be respected by pipeline
jobs.
The current implementation will not write to the log if a ConsoleLogFilter is not
serializable or missing the decorateLogger compatible with Run. This
would produce entries in the log for each step.


mwinter69:
obsolete as described in https://github.com/jenkinsci/workflow-job-plugin/pull/105

PR #60, opened 2018-05-18
Incrementalified

jglick:


jglick:
Mirror outage, retrying.

src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/StatusColumn.java
```diff
@@ -23,7 +22,7 @@ public FlowNodeViewColumn getDefaultInstance() {
 
         @Override
         public String getDisplayName() {
-            return Messages.StatusColumn_DisplayName();
+            return "Status";
```
carlossg: unrelated?

jglick: No, this is required in order for the plugin to be buildable after updating the parent POM.

PR #62, opened 2018-06-01
[JENKINS-51170] Enable the usage of the StepEnvironmentContributor extension point

t-8ch:
Depends on #60 and https://github.com/jenkinsci/workflow-step-api-plugin/pull/36/

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
@@ -68,14 +68,15 @@ public abstract class DefaultStepContext extends StepContext {
         if (key == EnvVars.class) {
             Run<?,?> run = get(Run.class);
             EnvironmentAction a = run == null ? null : run.getAction(EnvironmentAction.class);
-            EnvVars customEnvironment = a != null ? a.getEnvironment() : run.getEnvironment(getExecution().getOwner().getListener());
-            return key.cast(EnvironmentExpander.getEffectiveEnvironment(customEnvironment, (EnvVars) value, get(EnvironmentExpander.class)));
+            TaskListener taskListener = getListener(true);
+            EnvVars customEnvironment = a != null ? a.getEnvironment() : run.getEnvironment(taskListener);
+            return key.cast(EnvironmentExpander.getEffectiveEnvironment(customEnvironment, (EnvVars) value, get(EnvironmentExpander.class), this, taskListener));
         } else if (key == Launcher.class) {
             return key.cast(makeLauncher((Launcher) value));
         } else if (value != null) {
             return value;
         } else if (key == TaskListener.class) {
-            return key.cast(getListener());
+            return key.cast(getListener(false));
         } else if (Node.class.isAssignableFrom(key)) {
             Computer c = get(Computer.class);
             Node n = null;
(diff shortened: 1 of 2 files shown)


.mvn/extensions.xml
```diff
@@ -0,0 +1,7 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
```
jglick: BTW #60 has been merged, so to minimize your own PR’s diff, `git merge master` and push the merge here.

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -63,7 +63,7 @@
             Run<?,?> run = get(Run.class);
             EnvironmentAction a = run == null ? null : run.getAction(EnvironmentAction.class);
             EnvVars customEnvironment = a != null ? a.getEnvironment() : run.getEnvironment(get(TaskListener.class));
-            return key.cast(EnvironmentExpander.getEffectiveEnvironment(customEnvironment, (EnvVars) value, get(EnvironmentExpander.class)));
+            return key.cast(EnvironmentExpander.getEffectiveEnvironment(customEnvironment, (EnvVars) value, get(EnvironmentExpander.class), this, listener));
```
jglick: Do not use `listener` directly. Use `get(TaskListener.class)`.

t-8ch:
Implemented.

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -69,7 +69,7 @@
             Run<?,?> run = get(Run.class);
             EnvironmentAction a = run == null ? null : run.getAction(EnvironmentAction.class);
             EnvVars customEnvironment = a != null ? a.getEnvironment() : run.getEnvironment(getExecution().getOwner().getListener());
-            return key.cast(EnvironmentExpander.getEffectiveEnvironment(customEnvironment, (EnvVars) value, get(EnvironmentExpander.class)));
+            return key.cast(EnvironmentExpander.getEffectiveEnvironment(customEnvironment, (EnvVars) value, get(EnvironmentExpander.class), this, get(TaskListener.class)));
```
dwnusbaum: Appears to be causing test failures downstream in jenkinsci/workflow-cps-plugin#229 based on my local testing, see https://github.com/jenkinsci/workflow-cps-plugin/pull/229#pullrequestreview-193324238 for details.

jglick: Due to [this code](https://github.com/jenkinsci/workflow-cps-plugin/blob/a18ed9455cb9d83afa9d715448ec5e37f0802a1d/src/main/java/org/jenkinsci/plugins/workflow/cps/DSL.java#L237-L242), which I think ought to be fixed anyway: [JENKINS-47101](https://issues.jenkins-ci.org/browse/JENKINS-47101).

In the meantime, you could change this to directly call `getListener` but pass a new `boolean` param to let that return `TaskListener.NULL` in case `!node.isActive()`.

Or simply have `getListener` fall back to `getExecution().getOwner().getListener()` when called on an inactive node, so that any messages would be sent to the overall build log. This would probably be useful anyway in terms of solving some corner cases and race conditions.

t-8ch: @jglick Thanks!
I assume the same listener can be used the line above?

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -68,14 +68,15 @@
         if (key == EnvVars.class) {
             Run<?,?> run = get(Run.class);
             EnvironmentAction a = run == null ? null : run.getAction(EnvironmentAction.class);
-            EnvVars customEnvironment = a != null ? a.getEnvironment() : run.getEnvironment(getExecution().getOwner().getListener());
-            return key.cast(EnvironmentExpander.getEffectiveEnvironment(customEnvironment, (EnvVars) value, get(EnvironmentExpander.class)));
+            TaskListener taskListener = getListener(true);
+            EnvVars customEnvironment = a != null ? a.getEnvironment() : run.getEnvironment(taskListener);
```
jglick: This should be OK. See https://github.com/jenkinsci/workflow-support-plugin/commit/d5d1f46255b623587198a25f8c179c64f0b74d12: there was an issue using the per-step `TaskListener` which you also encountered via a different call stack. So now we just use the per-step listener if the step is running, and the general (build) listener otherwise.

PR #67, opened 2018-06-21
[JENKINS-52007] - Plugin POM 3.28: Update Animal Sniffer and other Maven plugins towards Java 11 support

oleg-nenashev:
Downstream of:

* https://github.com/jenkinsci/plugin-pom/pull/114
* https://github.com/mojohaus/animal-sniffer/pull/50

We need a release of Animal Sniffer to integrate it, but no hurry

CC @ndeloof @olamy @svanoort

PR #68, opened 2018-06-22
Java11 support - clean version

svanoort:
Clean version of #67 which pulls in the released versions. 

TODO:

- [x] Verify this is safe to run with Java 8 (no issues from the JBoss Marshalling bump)
- [ ] Test upgrading to this version with Pipelines running from old version
- [x] Add Enforcer exclusion for this JBoss Marshalling to allow use of multi-release JAR there.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,2 @@
-buildPlugin()
+//TODO: Currently some tests fail (JENKINS-52014, etc.) 
+buildPlugin(tests: [skip: true])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,2 +1,1 @@
-//TODO: Currently some tests fail (JENKINS-52014, etc.) 
-buildPlugin(tests: [skip: true])
+buildPlugin()
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-support</artifactId>
-    <version>2.18-SNAPSHOT</version>  <!-- Temp version to isolate work, remove before merge -->
+    <version>2.18</version>  <!-- Temp version to isolate work, remove before merge -->
     <packaging>hpi</packaging>
     <name>Pipeline: Supporting APIs</name>
     <url>https://wiki.jenkins.io/display/JENKINS/Pipeline+Supporting+APIs+Plugin</url>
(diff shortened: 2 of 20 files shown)


svanoort:
Still able to run Pipelines and start Jenkins with Java 8 on core 2.89.x -- as long as we revert the core dependency bump so it does not require 1.129 releases.  

This means the JBoss marshalling dependency is clean - barring enforcer grumbling about the multi-release JAR usage.

pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.12</version>
+        <version>3.16</version>
```
oleg-nenashev: Now you can update to 3.17 with newer Animal Sniffer

Jenkinsfile
```diff
@@ -1 +1,2 @@
-buildPlugin()
+//TODO: Currently some tests fail (JENKINS-52014, etc.) 
+buildPlugin(tests: [skip: true])
```
jglick: Perhaps better to `@Ignore` those so reviewers can see exactly what is still broken.

pom.xml
```diff
@@ -215,6 +222,25 @@
...
+            <plugin>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>display-info</id>
+                        <configuration>
+                            <rules>
+                                <enforceBytecodeVersion>
+                                    <excludes combine.children="append">
+                                        <!-- Prevents enforcer complaining about multi-release JAR -->
```
jglick: Where it the upstream bug?

svanoort:
@oleg-nenashev Note: the JBoss marshalling change is *not* back-compatible in the binary format -- that merits a serious upgrade warning (CompatibleSince, etc). 

Does not appear to have a memory leak, after tens of thousands of builds.

Build code, with build rotation set to keep last 100 builds. 

```

class Animal {
    String code;
    
    public Animal(String entry) {
        this.code = entry;
    }
}

for(int i=0; i<5; i++) {
    Animal an = new Animal("named"+i);
}

for (int i=0;i<5; i++) {
    echo 'cheese'
}
```

screen shot 2018-09-21 at 6 39 59 pm

oleg-nenashev:
Deployed the current version as `2.23-20181122.094059-1`

oleg-nenashev:
CC @jenkinsci/java11-support

PR #71, opened 2018-08-24
avoid printing too long argument in Pipeline Steps visualization

ArturHarasimiuk:
In some cases it is possible that node argument will be impossible to
wrap properly and is displayed as single line. In this case pipeline
steps view becomes hard to read because right most columns are moved
beyond the screen. This requires scrolling which is not so convenient.
Also status colum is outside of screen.

This change cuts arguments list to 80 characters which should be fine
for most of usages.

Signed-off-by: Artur Harasimiuk <artur.harasimiuk@intel.com>


src/main/java/org/jenkinsci/plugins/workflow/support/visualization/table/ArgumentsColumn.java
```diff
@@ -39,7 +40,7 @@
     @DataBoundConstructor public ArgumentsColumn() {}
 
     public String get(FlowNode node) {
-        return ArgumentsAction.getStepArgumentsAsString(node);
+        return StringUtils.substring(ArgumentsAction.getStepArgumentsAsString(node), 0, 80);
```
dwnusbaum: Maybe it would be better to truncate to 77, and if the text was truncated, add `...` to make it clear that the content was truncated?

ArturHarasimiuk: sure, no problem

PR #73, opened 2018-09-18
[WIP] Initial implementation for simple option for exposing build causes in a safe way

mikecirioli:
**WIP** **DO NOT MERGE**

One proposed approach to exposing build causes in a safe way for pipeline developers.  This approach depends on whitelisting the desired methods of any `Cause` implementations that you want to expose to pipeline developers ([see related PR for the script-security plugin which whitelists current implementations by default](https://github.com/jenkinsci/script-security-plugin/pull/225)).  

This approach has the advantage of allowing a user to find all `Cause`s that are a subclass of another `Cause`, as well as being very simple in its implementation.

Compare this approach against [this PR, which exposes @Exported fields as a `Map` of `Map`s using org.kohsuke.stapler.export.ModelBuilder](url)

**NOTE** linked PR's not yet created, will update shortly 

**Example pipeline Usage**
`pipeline {
    agent any

    stages {
        stage('debug') {
            steps {
                echo currentBuild.getBuildCauses().toString();
                script {
                    echo "---- test getting all causes --"
                    def causes = currentBuild.getBuildCauses()
                    echo "causes-->" + causes
                    echo "causes.size() -->" +causes.size()
                    for (cause in causes) {
                        echo "cause: " + cause.class
                        echo "short desc ->" + cause.getShortDescription()
                    }
                    
                    echo "---------- get specific cause types ----------"

                    causes = currentBuild.getBuildCauses(hudson.model.Cause$UserIdCause.class)
                    echo "causes-->" + causes
                    echo "causes.size() -->" +causes.size()
                    for (cause in causes) {
                        echo "cause: " + cause.class
                        echo "short desc ->" + cause.getShortDescription()
                    }

                    echo "------- get causes by superclass ------"
                    causes = currentBuild.getBuildCauses(hudson.model.Cause.class)
                    echo "causes-->" + causes
                    echo "causes.size() -->" +causes.size()
                    for (cause in causes) {
                        echo "cause: " + cause.class
                        echo "short desc ->" + cause.getShortDescription()
                    }
                }
            }
        }
        stage('declarative example') {
            when {
                expression {
                    currentBuild.getBuildCauses().get(0).getShortDescription() == 'Started by user anonymous'
                }
            }
            steps {
                echo 'we found our cause!'
            }
        }
    }
}

`

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -87,22 +87,24 @@ public final class RunWrapper implements Serializable {
 
     @Whitelisted
     public List<Cause> getBuildCauses() {
-        if (getRawBuild() != null) {
-            return Collections.unmodifiableList(getRawBuild().getCauses());
+        Run<?,?> run = getRawBuild();
+        if (run != null) {
+            return Collections.unmodifiableList(run.getCauses());
         }
         else {
-            return new ArrayList<Cause>();
+            return Collections.emptyList();
         }
     }
 
     @Whitelisted
     public List<Cause> getBuildCauses(Class<Cause> causeType) {
-        if (getRawBuild() != null) {
+        Run<?,?> run = getRawBuild();
+        if (run != null) {
             return Collections.unmodifiableList(
-                    getRawBuild().getCauses().stream().filter(cause -> causeType.isInstance(cause)).collect(Collectors.toList()));
+                    run.getCauses().stream().filter(cause -> causeType.isInstance(cause)).collect(Collectors.toList()));
         }
         else {
-            return new ArrayList<Cause>();
+            return Collections.emptyList();
         }
     }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -41,6 +41,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
@@ -86,14 +87,13 @@ public final class RunWrapper implements Serializable {
 
     @Whitelisted
     public List<Cause> getBuildCauses() {
-        return Collections.unmodifiableList(getRawBuild().getCauses());
+        return Collections.unmodifiableList(Objects.requireNonNull(getRawBuild()).getCauses());
     }
 
     @Whitelisted
     public List<Cause> getBuildCauses(Class<Cause> causeType) {
-        return Collections.unmodifiableList(getRawBuild().getCauses().stream().filter(cause -> causeType.isInstance(cause)
-        ).collect
-                (Collectors.toList()));
+        return Collections.unmodifiableList(Objects.requireNonNull(getRawBuild().getCauses()).stream().filter(cause -> causeType
+                .isInstance(cause)).collect(Collectors.toList()));
     }
 
     @Whitelisted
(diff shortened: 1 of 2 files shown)


dwnusbaum:
Is this actually safe? What if the `Cause`s aren't serializable?

jtnord:
@dwnusbaum good point - but then what about all those SCM [`ChangeLogSets`](https://javadoc.jenkins.io/hudson/scm/ChangeLogSet.html)  that are exposed in [`RunWrapper`](https://github.com/jenkinsci/workflow-support-plugin/blob/8105510ba5617c19e90c8cf5ee467fb559833d8d/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java) as well?

Is that an existing bug, or does the workflow plugin do some magic with xstream rather than java serialization?

dwnusbaum:
@jtnord Based on a quick test, it looks like JBoss marshalling is used to save the program state:

    Caused: java.io.NotSerializableException: java.lang.Thread
        at org.jboss.marshalling.river.RiverMarshaller.doWriteObject(RiverMarshaller.java:860)
        at org.jboss.marshalling.river.RiverMarshaller.doWriteFields(RiverMarshaller.java:1032)
        at org.jboss.marshalling.river.RiverMarshaller.doWriteSerializableObject(RiverMarshaller.java:988)
        at org.jboss.marshalling.river.RiverMarshaller.doWriteObject(RiverMarshaller.java:854)
        at org.jboss.marshalling.river.BlockMarshaller.doWriteObject(BlockMarshaller.java:65)
        at org.jboss.marshalling.river.BlockMarshaller.writeObject(BlockMarshaller.java:56)
        at org.jboss.marshalling.MarshallerObjectOutputStream.writeObjectOverride(MarshallerObjectOutputStream.java:50)
        at org.jboss.marshalling.river.RiverObjectOutputStream.writeObjectOverride(RiverObjectOutputStream.java:179)
        at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:344)
        at java.util.HashMap.internalWriteEntries(HashMap.java:1790)
        at java.util.HashMap.writeObject(HashMap.java:1363)
        at sun.reflect.GeneratedMethodAccessor144.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.jboss.marshalling.reflect.SerializableClass.callWriteObject(SerializableClass.java:273)

It looks that library doesn't require classes to implement the `Serializable` marker interface, so I think the only issue would be if there are `Cause` implementations in use with non-serializable fields. The implementations in Jenkins core look ok, but I didn't look any further than that.

dwnusbaum:
Note that the `changeSet` help text mentions that you should only use it inside of an `@NonCPS` method: https://github.com/jenkinsci/workflow-support-plugin/blob/8105510ba5617c19e90c8cf5ee467fb559833d8d/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper/help.html#L21

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -81,6 +85,27 @@ public RunWrapper(Run<?,?> build, boolean currentBuild) {
...
+    }
+
+    @Whitelisted
+    public List<Cause> getBuildCauses(Class<Cause> causeType) {
+        if (getRawBuild() != null) {
+            return Collections.unmodifiableList(
+                    getRawBuild().getCauses().stream().filter(cause -> causeType.isInstance(cause)).collect(Collectors.toList()));
+        }
+        else {
+            return new ArrayList<Cause>();
```
joseblas: you could use immutable list too: 

> Collections.<Cause>emptyList()

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -81,6 +85,27 @@ public RunWrapper(Run<?,?> build, boolean currentBuild) {
...
+        }
+    }
+
+    @Whitelisted
+    public List<Cause> getBuildCauses(Class<Cause> causeType) {
+        if (getRawBuild() != null) {
+            return Collections.unmodifiableList(
+                    getRawBuild().getCauses().stream().filter(cause -> causeType.isInstance(cause)).collect(Collectors.toList()));
```
joseblas: probably I'm a bit obsessed, but we could use only once _getRawBuild().getCauses()_

mikecirioli: agreed, it bothers me too.  Will fix

PR #74, opened 2018-09-21
[JENKINS-26138] Display all workspaces for a build

jglick:
[JENKINS-26138](https://issues.jenkins-ci.org/browse/JENKINS-26138)

screenshot

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceRunAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceRunAction.java
@@ -1,0 +1,96 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2018 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceRunAction.java
```diff
@@ -0,0 +1,96 @@
...
+        return "Workspaces";
+    }
+
+    @Override public String getUrlName() {
+        return "ws";
+    }
+
+    public List<WorkspaceActionImpl> getActions() throws Exception {
+        List<WorkspaceActionImpl> r = new ArrayList<>();
+        for (FlowNode node : new DepthFirstScanner().allNodes(owner.get())) {
```
svanoort: 🐜  Can be done with new `DepthFirstScanner().filteredNodes(owner.get(), FlowScanningUtils.hasActionPredicate(WorkspaceActionImpl.class))` as a one-liner - then you iterate over that because it's an Iterable

svanoort: Might consider caching

jglick: I actually started off using `hasActionPredicate` just like you said, but it does not really help anything since I still need to call `getAction` or similar and check for null. This is shorter and should be just as fast.

Caching is neither desirable nor necessary here: it is called once per page load, which is only when a user actually clicks the **Workspaces** link, and the result can change during a running build.

svanoort: `hasActionPredicate` does the null check for you :)

svanoort: 🐛 I *think* you want to do `getOrNull()` here for UI display - so it doesn't block waiting for loading (better to just wait for a page refresh), and do a nullCheck - `get()` may cause the thread to be blocked a long time.

jglick: I would rather have the thread blocked momentarily than potentially have no information.

src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceRunAction.java
```diff
@@ -0,0 +1,110 @@
...
+            exec = owner.get();
+        } catch (IOException x) {
+            LOGGER.log(Level.WARNING, null, x);
+            // Broken flow, cannot display anything.
+            return Collections.emptyList();
+        }
+        List<WorkspaceActionImpl> r = new ArrayList<>();
+        for (FlowNode node : new DepthFirstScanner().allNodes(exec)) {
+            for (WorkspaceActionImpl action : node.getActions(WorkspaceActionImpl.class)) {
+                r.add(action);
```
dwnusbaum: If you have 2 separate `node` blocks that use the same agent, you'll end up with two `WorkspaceActionImpl` instances that have unique nodes but identical workspaces, so there will be 2 links on the page with identical text.  Kind of awkward, but the current behavior seems fine to me.

PR #78, opened 2018-10-24
[JENKINS-54227] Safely expose the Cause(s) associated with the current build

mikecirioli:
Use stapler's `ModelBuilder` to create a JSON representation of a `Cause`s `@Exported` fields:

For example, a build with a `hudson.model.Cause$UserId` cause produce the following output:
 
```
[{ "_class":"hudson.model.Cause$UserIdCause", "shortDescription":"Started by user anonymous", "userId":"tester", "userName":"anonymous" }]
```

The JSON objects in the resulting array can be used directly in a pipeline:

```
assert currentBuild.getBuildCauses().size() == 1
assert currentBuild.getBuildCauses()[0].userId == 'tester'
echo currentBuild.getBuildCauses()[0].shortDescription
```

Additionally, you can filter the result of `currentBuild.getBuildCauses()` by passing a class (or superclass) of the type you would like to filter by.  For example, to get a list of build `Cause`s that only contains `Cause`s of type `hudson.model.Cause$UserIdCause`, call the method like this:

`
echo currentBuild.getBuildCauses(hudson.model.Cause$UserIdCause).size()
`
Calling the method with `hudson.model.Cause` as a parameter will return all the `Cause`s associated with a build, since all `Cause`s are a sub-class of this type.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.22</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.121.1</jenkins.version>
+        <jenkins.version>2.130</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.22</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.130</jenkins.version>
+        <jenkins.version>2.121.1</jenkins.version>
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -150,17 +150,28 @@ public final class RunWrapper implements Serializable {
         }
     }
 
+    /**
+     * Filters the returned list by the type of <code>Cause</code> class passed as input
+     * ex. <code>getBuildCauess('hudson.model.Cause$UserIdCause')</code> would return only
+     * <code>Cause</code>s of that type
+     *
+     * @param className A string containing the fully qualified name for the class type to filter the result list by
+     * @return a <code>JSONArray</code> of <code>Cause</code>s of the specified type
+     * @throws IOException
+     */
     @Whitelisted
-    public JSONArray getBuildCauses(Class<Cause> clazz) throws IOException {
+    public JSONArray getBuildCauses(String className) throws IOException, ClassNotFoundException {
+        Class clazz = Class.forName(className);
         JSONArray result = new JSONArray();
-        StringWriter w = new StringWriter();
 
         for(Cause cause : build().getCauses()) {
             if (clazz.isInstance(cause)) {
+                StringWriter w = new StringWriter();
                 CauseAction causeAction = new CauseAction(cause);
                 DataWriter writer = JSON.createDataWriter(causeAction, w);
                 Model<CauseAction> model = new ModelBuilder().get(CauseAction.class);
                 model.writeTo(causeAction, writer);
+                // return a slightlly cleaner object by removing the outer object
                 result.add(JSONObject.fromObject(w.toString()).getJSONArray("causes").get(0));
             }
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -134,20 +134,8 @@ public final class RunWrapper implements Serializable {
     }
 
     @Whitelisted
-    public JSONArray getBuildCauses() throws IOException {
-        CauseAction causeAction = build().getAction(CauseAction.class);
-
-        if (causeAction != null) {
-            StringWriter w = new StringWriter();
-            DataWriter writer = JSON.createDataWriter(causeAction, w);
-            Model<CauseAction> model = new ModelBuilder().get(CauseAction.class);
-            model.writeTo(causeAction, writer);
-                JSONObject result =  JSONObject.fromObject(w.toString());
-                // return a slightlly cleaner object
-                return result.getJSONArray("causes");
-        } else {
-            return new JSONArray();
-        }
+    public JSONArray getBuildCauses() throws IOException, ClassNotFoundException {
+        return getBuildCauses("hudson.model.Cause");
     }
 
     /**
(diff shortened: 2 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -125,6 +133,40 @@ public int getNumber() throws AbortException {
...
+                JSONObject result =  JSONObject.fromObject(w.toString());
+                // return a slightlly cleaner object
+                return result.getJSONArray("causes");
+        } else {
+            return new JSONArray();
+        }
+    }
+
+    @Whitelisted
+    public JSONArray getBuildCauses(Class<Cause> clazz) throws IOException {
```
dwnusbaum: Regardless of the future compatibility issues with RunWrapper, I don't think we want anyone to be referring to Jenkins classes directly in their pipeline. Maybe make the parameter a `String` so users can pass the class name directly. I guess we could make this nicer for users by adding `@Symbol` annotations to `Cause` classes and then allowing users to pass those symbol names instead of the full class name.

mikecirioli:
You are correct, I took this approach as I have limited time and this feature has been a fairly high priority ask.  I am open to possibly converting this to a new pipeline step in order to reducing dependencies on the workflow support plugin, my primary goal was to provide a simple way to access build causes.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
cyrille-leclerc: @mikecirioli this method doesn't seem to be needed in the UX we have defined, I propose to remove this signature for the moment.

jtnord: well you want to get causes that are a `X` or it's subclasses.  You can not do that in pipeline without a big mess.  Possibly it should not take a class but a classname or something.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -125,6 +133,40 @@ public int getNumber() throws AbortException {
...
+            model.writeTo(causeAction, writer);
+                JSONObject result =  JSONObject.fromObject(w.toString());
+                // return a slightlly cleaner object
+                return result.getJSONArray("causes");
+        } else {
+            return new JSONArray();
+        }
+    }
+
+    @Whitelisted
```
jtnord: should document what the clazz parameter does.
ie it filters to return only causes that are assignable to type `clazz`  (can this be marked as beta in the help somehow to allow people to actually have somethign now but can be done better latter?)

pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.22</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.121.1</jenkins.version>
+        <jenkins.version>2.130</jenkins.version>
```
jtnord: is the jenkins bump actually needed?

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper/help.html
```diff
@@ -1,4 +1,6 @@
 <dl>
+    <dt><code>getBuildCauses</code></dt><dd>Returns a JSON array of build causes for the current build</dd>
+    <dt><code>getBuildCauses(Class&lt;Cause&gt;)</code></dt><dd>Returns a <code>JSON array</code> of build causes of type <code>Cause</code> for the current build, or an empty <code>JSON array</code> if no causes of the specified type apply to the current build</code></dd>
```
jtnord: iff we are thinking of changing this API can you mark it as BETA or EXPERIMENTAL.

pom.xml
mikecirioli: i recall needing to do that to make my local testing work, i will take a look and see if it is really needed outside that context.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper/help.html
```diff
@@ -1,4 +1,6 @@
 <dl>
+    <dt><code>getBuildCauses</code></dt><dd>Returns a JSON array of build causes for the current build</dd>
+    <dt><b>EXPERIMENTAL - MAY CHANGE</b>  <code>getBuildCauses(String causeClass)</code></dt><dd>Takes a string representing the fully qualified <code>Cause</code> class and returns a <code>JSON array</code> of build causes filtered by that type for the current build, or an empty <code>JSON array</code> if no causes of the specified type apply to the current build</code></dd>
```
svanoort: Sounds a lot like the Beta annotation

mikecirioli: Is there a preferred wording @svanoort ?  I did some quick searches but couldn't find anything that  looked like a pattern.

mikecirioli:
@dwnusbaum @svanoort @cyrille-leclerc  - is there any accompanying product documentation that I should update in conjunction with this PR?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -125,6 +133,39 @@ public int getNumber() throws AbortException {
...
+     * Filters the returned list by the type of <code>Cause</code> class passed as input
+     * ex. <code>getBuildCauess('hudson.model.Cause$UserIdCause')</code> would return only
+     * <code>Cause</code>s of that type
+     *
+     * @param className A string containing the fully qualified name for the class type to filter the result list by
+     * @return a <code>JSONArray</code> of <code>Cause</code>s of the specified type
+     * @throws IOException
+     */
+    @Whitelisted
+    public JSONArray getBuildCauses(String className) throws IOException, ClassNotFoundException {
```
svanoort: Accepting a String classname directly.... eh, not the prettiest thing (oof, that ClassNotFoundException!) but I can't think of another way we might accomplish this.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper/help.html
svanoort: No, no, I mean on the API itself in the Java code -- but it's not surfaced for users in such a way (more of a purely-developer-side thing).

svanoort: Anyway, no chance needed to the code here

mikecirioli: oh yeah, i saw that but felt for this API it wasn't needed as this will almost surely only be called from pipeline scripts

mikecirioli:
@dwnusbaum @svanoort @abayer please merge at your convenience - I do not have sufficient power to do it myself it seems.  thanks!

PR #79, opened 2018-10-30
Small refactor for readability, added new test that trows error

LavaGolem:
Added new test that check exception

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/concurrent/TimeoutTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/concurrent/TimeoutTest.java
@@ -44,7 +44,6 @@ public class TimeoutTest {
         try (Timeout timeout = Timeout.limit(5, TimeUnit.SECONDS)) {
             Thread.sleep(1_000);
         }
-        Thread.sleep(5_000);
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/concurrent/TimeoutTest.java
```diff
@@ -54,7 +53,14 @@
         } catch (InterruptedException x) {
             // good
         }
-        Thread.sleep(1_000);
+    }
+
+    @Test(expected = InterruptedException.class)
+    public void testInterruptedException() throws InterruptedException {
```
dwnusbaum: Isn't this identical to [`TimeoutTest#failed`](https://github.com/jenkinsci/workflow-support-plugin/blob/e2ba3ca65dfe057aeb8986b7e1f114051e513264/src/test/java/org/jenkinsci/plugins/workflow/support/concurrent/TimeoutTest.java#L49) other than using `@Test(expected = ...)` instead of manually catching the exception? It doesn't seem useful to have both tests, so it seems better not to add this one.

dwnusbaum:
I am going to go ahead and close this since the test appears to be redundant, but please feel free to reopen it if you add any tests for currently untested functionality.

PR #82, opened 2018-11-23
Prepare an experimental release with Java 11 compatibility

oleg-nenashev:
WiP branch for Java 11 support in a Java11-only experimental release. Based on #68 from @svanoort , but it will include more fixes.

Generally we may end up with landing it as a `java11-support` branch, to be discussed

This change cannot be released until https://github.com/jenkins-infra/update-center2/pull/253 is integrated and deployed. Otherwise it will impact users of the common experimental update center

@jenkinsci/java11-support


batmat:
Also deployed as a SNAPSHOT `3.0-java11-alpha-1-20181203.134713-1` for places where it's not possible/easy to consume Incrementals.

pom.xml
```diff
@@ -95,9 +96,16 @@
             <version>1.39</version>
         </dependency>
         <dependency>
+            <!-- Requires Java 9+ -->
```
oleg-nenashev:
```suggestion
            <!-- Required for running with Java 9+ -->
```

oleg-nenashev: Not really needed anymore, devtools have been updated

oleg-nenashev:
`Minimum-Java-Version` is injected correctly

screenshot 2018-12-05 at 13 28 02

pom.xml
```diff
@@ -96,7 +96,7 @@
             <version>1.39</version>
         </dependency>
         <dependency>
-            <!-- Requires Java 9+ -->
+            <!-- Required for running with Java 9+ -->
```
batmat: LGTM. I suppose this is fine, despite the fact we don't anyway really support Java 9, but 8, or 11+ and nothing in between.

PR #83, opened 2018-12-06
getBuildCauses(...) now uses the uber classloader

mikecirioli:
getBuildCauses(...) now uses the uber classloader so that classes provided by other extension points can be found.  Also provides sane error handling when a CNF exception does happen (logs a warning and returns an empty JSONArray)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -42,9 +42,12 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 
+import jenkins.model.Jenkins;
 import jenkins.scm.RunWithSCM;
 import net.sf.json.JSONArray;
 import net.sf.json.JSONObject;
@@ -69,6 +72,8 @@ public final class RunWrapper implements Serializable {
 
     private final String externalizableId;
     private final boolean currentBuild;
+
+    private static final Logger LOGGER = Logger.getLogger(RunWrapper.class.getName());
 
     public RunWrapper(Run<?,?> build, boolean currentBuild) {
         this.externalizableId = build.getExternalizableId();
@@ -152,8 +157,16 @@ public final class RunWrapper implements Serializable {
      */
     @Whitelisted
     public JSONArray getBuildCauses(String className) throws IOException, ClassNotFoundException {
-        Class clazz = Class.forName(className);
+        final ClassLoader cl = Jenkins.get().pluginManager.uberClassLoader;
         JSONArray result = new JSONArray();
+        final Class clazz;
+
+        try {
+            clazz = cl.loadClass(className);
+        } catch (ClassNotFoundException cnfe) {
+            LOGGER.log(Level.WARNING, "No class found for {0}", className);
+            return result;
+        }
 
         for(Cause cause : build().getCauses()) {
             if (clazz.isInstance(cause)) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -169,6 +175,21 @@ public JSONArray getBuildCauses(String className) throws IOException, ClassNotFo
         return result;
     }
 
+    private boolean isClassOrSubclass(Class<?> clazz, String classNameToCheck) {
+        String causeClassName = clazz.getName();
+        if (causeClassName.equals("java.lang.Object")) {
+            return false;
```
jtnord: `return causeClassName.equals(classNameToCheck)`  incase the user enters "java.lang.Object")

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -169,6 +175,21 @@ public JSONArray getBuildCauses(String className) throws IOException, ClassNotFo
...
+        } else if (causeClassName.equals(classNameToCheck)) {
+            return true;
+        } else {
+            if (causeClassName.contains("$")) {
```
jtnord: if the cause is foo.bar.Wibble$Cause would you not expect the user to enter that?  Are there any instances of this, worth writing a test for?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
mikecirioli: And also to exit the recursion...

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
mikecirioli: I'll look into writing a test but i think its already covered. Take the case of `hudson.build.Cause$UserCause` - if the user wants to check to see if a cause is of that type, they have to enter that string.  I found that `clazz.getClass().getSuperClass().getName()` was returning `java.lang.Object` when it was passed to the next level of recursion (which makes sense i believe) which is why i added the check to pass `clazz.getEnclosingClass()` when needed.

dwnusbaum: I'm also confused by this check. It seems like you'd never want to check the enclosing class, since that is just about source-level organization, not about the actual sub/superclass relationships which should be handled by the recursive calls to `getSuperclass()`. 

> It is a bit odd to me though, because class.getClass().getSuperClass().getName() returns the enclosing class when evaluated in the ide


In the case of `hudson.build.Cause$UserCause`, `UserCause` directly extends `hudson.build.Cause`, so `hudson.build.Cause` is both the superclass and the enclosing class. Is that what was happening?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
dwnusbaum: `Object.class.getSuperclass` [will return null](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getSuperclass--), so maybe easier to use that as the base case and handle `Object` with the next `if` just like other classes.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -169,6 +175,21 @@ public JSONArray getBuildCauses(String className) throws IOException, ClassNotFo
...
+                return isClassOrSubclass(clazz.getEnclosingClass(), classNameToCheck);
+            } else {
+                return isClassOrSubclass(clazz.getClass().getSuperclass(), classNameToCheck);
```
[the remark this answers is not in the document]
mikecirioli: I am starting to think that this has become overly complex, and that the most likely use case would be for checking for an exact match (ie. show me only causes of type `hudson.model.Cause$UserIdCause`)

PR #84, opened 2018-12-07
Java11 Support

svanoort:
For experimental release (very limited scope) -- experimental update center ONLY, limited to only Jenkins running Java 11.


oleg-nenashev:
unlikely, but it's still nice to have it to track the changes. I used the same approach with Java 11 support branch for the core. You are done when the diff is zero, or something like that

svanoort:
@batmat Probably adapt it and merge it once the final kinds for Java 11 are worked out

oleg-nenashev:
I confirm the correct deployment on the experimental UCs

screenshot 2018-12-08 at 00 07 36

PR #85, opened 2018-12-10
Fixes getBuildCause(String classname) so that it no longer relies on class loading

mikecirioli:
Fixes getBuildCause(String classname) so that it no longer relies on class loading to determine if a class meets the filter criteria.  This eliminates possible CNFE's when checking for build causes supplied by other @Extension points.

This PR supersedes #83


src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -138,7 +138,18 @@ public int getNumber() throws AbortException {
...
-        return getBuildCauses("hudson.model.Cause");
+        JSONArray result = new JSONArray();
+
+        for(Cause cause : build().getCauses()) {
+            StringWriter w = new StringWriter();
+            CauseAction causeAction = new CauseAction(cause);
+            DataWriter writer = JSON.createDataWriter(causeAction, w);
+            Model<CauseAction> model = new ModelBuilder().get(CauseAction.class);
+            model.writeTo(causeAction, writer);
+            // return a slightlly cleaner object by removing the outer object
```
jglick: typo

PR #86, opened 2018-12-27
[JENKINS-55174] Update JBoss marshalling and unignore fixed tests

dwnusbaum:
See [JENKINS-55174](https://issues.jenkins-ci.org/browse/JENKINS-55174). Based against #84.

It looks like the root cause of the NPE we were seeing was [JBMAR-221](https://issues.jboss.org/browse/JBMAR-221), which was fixed in JBoss Marshalling 2.0.6.Final, so I've updated to that version and unignored the previously-failing tests.

CC @jenkinsci/java11-support

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -99,13 +99,7 @@
             <!-- Required for running with Java 9+ -->
             <groupId>org.jboss.marshalling</groupId>
             <artifactId>jboss-marshalling-river</artifactId>
-            <version>2.0.5.Final</version>
-        </dependency>
-        <dependency>
-            <groupId>org.jenkins-ci</groupId>
-            <artifactId>annotation-indexer</artifactId>
-            <version>1.12</version>
-            <scope>test</scope>
+            <version>2.0.6.Final</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ThrowablePickleTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ThrowablePickleTest.java
@@ -47,7 +47,6 @@ public class ThrowablePickleTest {
     @Rule public LoggerRule logging = new LoggerRule().record(ThrowablePickle.class, Level.FINE);
 
     @Issue("JENKINS-51390")
-    @Ignore("Still has problems with new JBoss Marshalling, temporarily disabled while debugging.")
     @Test public void smokes() throws Exception {
         String beName = BadException.class.getName();
         rr.then(r -> {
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -99,13 +99,7 @@
             <!-- Required for running with Java 9+ -->
             <groupId>org.jboss.marshalling</groupId>
             <artifactId>jboss-marshalling-river</artifactId>
-            <version>2.0.5.Final</version>
-        </dependency>
-        <dependency>
-            <groupId>org.jenkins-ci</groupId>
-            <artifactId>annotation-indexer</artifactId>
```
dwnusbaum: `mvn dependency:tree` shows that we already pick up `org.jenkins-ci:annotation-indexer:jar:1.12:provided` through core, so better to remove this.

batmat:
Thanks a lot Devin!

dwnusbaum:
@MarkEWaite The wire format of JBoss marshalling is not guaranteed to be backwards compatible across version changes, so that might just have been an issue across the restart where you upgraded the plugin. Do you still see that issue on restarts for in-progress builds even when using only the same version of the plugin across restarts?

MarkEWaite:
Yes @dwnusbaum I see the same issue when restarting with no change to plugin version.  I've been through several restart cycles to see if the failures are consistent.  

I've now changed the consistency setting for maximum consistency (worst performance) and will try a restart with jobs in progress.  I'll check the status after the restart and let you know the results

MarkEWaite:
It appears that `buildPlugin` sets the consistency to performance optimized, which overrides my attempt to set it for the entire Jenkins instance.  I'll need to find a long-running Pipeline job that is not building a plugin before I can correctly test with maximum consistency.

In the interim, shall I open a Jira ticket for the new behavior so that we can track it in a Jira ticket rather than in comments on this pull request?

dwnusbaum:
@MarkEWaite Yeah I'd go ahead and open a ticket, and please include any logs/error messages that occur at the time of the failure. FWIW, I have not been able to reproduce any issues across restarts under any durability setting and the latest Pipeline-related plugins, although my Pipelines are relatively simple and I am running Jenkins using Java 8.

pom.xml
alecharp: I do understand the point, but I disagree. If you use a library, you should say so in your `pom.xml`. See `mvn dependency:analyze`, you will have "undeclared" libraries. The point is, if you don't list the dependency, and the other one which was providing it by transitivity stop using it (won't happen with core, yes), then you might have issues during an upgrade.

batmat:
> Merging into the java11 branch now, would suggest we cut a release of this tomorrow

Yes, please! :-) Fully agree. Having a release for the plugin that fixes issue and moves the ball forward does make perfect sense.

pom.xml
dwnusbaum: Yes, my opinion would be different if the library was not in core, but depending on a specific version of the library here just opens us up to accidental version mismatches with core which could be problematic.

PR #88, opened 2019-01-14
Remove dependency on obsolete workflow-scm-step tests artifact

dwnusbaum:
Fixes PCT failures when running against workflow-scm-step 2.7. As of jenkinsci/workflow-scm-step-plugin#17, the workflow-scm-steps tests artifact is no longer produced, because the useful classes it had were moved to scm-api in jenkinsci/workflow-scm-step-plugin#15 (landed originally in workflow-scm-step-2.5, i.e. the tests artifact has been obsolete for a long time).

CC @halkeye

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
         <git-plugin.version>3.7.0</git-plugin.version>
-        <workflow-scm-step-plugin.version>2.6</workflow-scm-step-plugin.version>
+        <workflow-scm-step-plugin.version>2.7</workflow-scm-step-plugin.version>
         <workflow-step-api-plugin.version>2.13</workflow-step-api-plugin.version>
         <scm-api-plugin.version>2.2.6</scm-api-plugin.version>
     </properties>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -69,7 +69,7 @@
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
         <git-plugin.version>3.7.0</git-plugin.version>
-        <workflow-scm-step-plugin.version>2.6</workflow-scm-step-plugin.version>
+        <workflow-scm-step-plugin.version>2.7</workflow-scm-step-plugin.version>
```
dwnusbaum: Strictly speaking, there is no need to update this to fix the issue, but we might as well bump to the lastest version while we're here so we can catch these kinds of issues before the PCT.

halkeye:
awesome~~~

thanks for looking into this so quickly

batmat:
Ping @jenkinsci/java11-support for more eyes.

dwnusbaum:
@batmat Do you need this urgently? I am happy to merge+release now, but am not aware of a pressing need for it so it is on my TODO list.

batmat:
Afaik we had nothing, at least yet, tracking this.
I just guess that the sooner, without urgency, the better so we avoid potential time spent running PCT and failing on this.

Should likely add it in https://docs.google.com/document/d/1oluVrNVpQhXCIwW9CYVm09Y1vPc3H77d3q92LrzcpDw/edit?usp=drivesdk for completeness.

Thanks!

dwnusbaum:
Went ahead and released 3.1 since I had some time.

PR #92, opened 2019-03-25
Java 11 readiness: build using recommended configurations

batmat:
Using the right variations of Jenkins versions, JDK 8 & 11, Windows/Linux...

For Java 11 readiness, apply recommended configuration:
https://wiki.jenkins.io/display/JENKINS/Java+11+Developer+Guidelines#Java11DeveloperGuidelines-MakesureyourpluginistestedinContinuousIntegrationonJava8andJava11atthesametime

And fixing what was revealed through this.

@jenkinsci/java11-support

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
@@ -157,7 +157,7 @@ public class LogActionImpl extends LogAction implements FlowNodeAction, Persiste
     }
 
     /**
-     * Used from <tt>console.jelly</tt> to write annotated log to the given output.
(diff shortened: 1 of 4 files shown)


batmat:
@dwnusbaum could you please consider merging this? Thanks!

src/main/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImpl.java
```diff
@@ -157,7 +157,7 @@ private File getLogFile() throws IOException {
     }
 
     /**
-     * Used from <tt>console.jelly</tt> to write annotated log to the given output.
+     * Used from <code>console.jelly</code> to write annotated log to the given output.
```
jglick: `{@code …}` is preferred BTW.

PR #93, opened 2019-03-26
Add i18n support for workspace

LinuxSuRen:


src/main/resources/org/jenkinsci/plugins/workflow/support/actions/WorkspaceRunAction/index.properties
```diff
@@ -0,0 +1,2 @@
+Workspaces=Workspaces
+Workspaces\ for\ =Workspaces for {0}
```
dwnusbaum: nit: Maybe better to switch to a key with no spaces like `WorkspacesFor` to avoid any confusion/complications.

src/main/resources/org/jenkinsci/plugins/workflow/support/actions/Messages.properties
```diff
@@ -0,0 +1 @@
+Workspaces=Workspaces
```
dwnusbaum: Findbugs wants these to start with a lowercase letter (I always forget this myself as well):

> The method name org.jenkinsci.plugins.workflow.support.actions.Messages.Workspaces() doesn't start with a lower case letter [org.jenkinsci.plugins.workflow.support.actions.Messages] At Messages.java:[line 35] NM_METHOD_NAMING_CONVENTION

Actually though, should this file be deleted? It doesn't look like it's used anywhere.

LinuxSuRen:
Looks like the error is not related.

dwnusbaum:
> javadoc: error - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.

I think you can fix the error by updating to [version 3.43 of the parent POM or newer](https://github.com/jenkinsci/plugin-pom/releases).

dwnusbaum:
@LinuxSuRen I think you reserved the merge conflict the wrong way in c6df9f2, we want to go to 3.43+, but right now it is still at 3.42. You might want to merge with master first in case the conflict came from your branch being based against an old commit on master.

PR #95, opened 2019-03-28
[JENKINS-41854] DynamicContext

jglick:
Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/42.


jglick:
Putting this on hold while I check whether it is feasible to interpret `DynamicContext` from `ContextVariableSet` as discussed in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/101#discussion_r289485748.

jglick:
Have a better solution.

PR #97, opened 2019-06-15
fix name error

hummerstudio:
This plugin is not an API, I think it's name is not correct.


hummerstudio:
@dwnusbaum OK. Thanks for the reply. Maybe we can add a README.md for this to avoid misunderstanding.

dwnusbaum:
I added a README in #103, so I'm closing this. Thanks!

PR #98, opened 2019-07-21
[JENKINS-58586] Implement RunWrapper getProjectDescription() and setP…

stuartrowe:
…rojectDescription() methods

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
@@ -120,7 +120,17 @@ public final class RunWrapper implements Serializable {
             build().setDisplayName(n);
         }
     }
-	
+
+    @Whitelisted
+    public void setProjectDescription(String d) throws IOException {
+        if (!currentBuild) {
+            throw new SecurityException("can only set the description on the project of the current build");
+        }
+        try (ACLContext ctx = ACL.as(ACL.SYSTEM)) {
+            build().getParent().setDescription(d);
+        }
+    }
+
     @Whitelisted
     public void setKeepLog(boolean b) throws IOException {
         if (!currentBuild) {
@@ -259,6 +269,11 @@ public final class RunWrapper implements Serializable {
     @Whitelisted
     public String getFullProjectName() throws AbortException {
         return build().getParent().getFullName();
+    }
+
+    @Whitelisted
+    public String getProjectDescription() throws AbortException {
+        return build().getParent().getDescription();
     }
 
     @Whitelisted
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.42</version>
+        <version>3.43</version>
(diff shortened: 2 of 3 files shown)


stuartrowe:
See [JENKINS-58586](https://issues.jenkins-ci.org/browse/JENKINS-58586).

@dwnusbaum please review.

dwnusbaum:
Build failure is not your fault, but you can fix it by updating from version 3.42 of the parent to 3.43+, I think the current version is [3.47](https://github.com/jenkinsci/plugin-pom/releases).

> The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.

stuartrowe:
@dwnusbaum I included your suggested fix to bump the parent version. Is it alright for that fix to be included in this PR even though it's not related to the JIRA issue being addressed?

stuartrowe:
@dwnusbaum 👈

src/main/java/org/jenkinsci/plugins/workflow/support/steps/build/RunWrapper.java
```diff
@@ -120,7 +120,17 @@ public void setDisplayName(String n) throws IOException {
             build().setDisplayName(n);
         }
     }
-	
+
+    @Whitelisted
+    public void setProjectDescription(String d) throws IOException {
+        if (!currentBuild) {
```
dwnusbaum: I'm not sure about this. It makes sense that we let a Pipeline change the description/name of its own build, but changing the description of its job, which is shared across all builds, seems different. What is your use case? 

Unfortunately I think more granular access control here would require that something like the authorize-project plugin be installed and configured, so this is basically the best we can do.

stuartrowe: I understand your position and I could always use JobDSL or similar to do this. Since I already have a workaround using a `@NonCPS` method in my shared library ,I don't really need this to get merged.

dwnusbaum: FWIW I was mostly just curious to understand how you intended to use it; do you have a job that creates other jobs or something?

stuartrowe: A bit of background: my shared library defines classes for my various build types. I am using Jenkinsfiles as super-config files that create a build object with the given config and then runs it. I've added a config for the project description to some of these Jenkinsfiles so that when the build runs, the project description is updated if the config changes.

As mentioned earlier, I'm currently using a `@NonCPS` method to do this. That method accesses the project instance through `currentBuild.getRawBuild().getParent()` and then updates the description. I wanted to replace that with a single pipeline step and get rid of the `@NonCPS` method.

dwnusbaum: Ok, what you are doing in your shared library seems like a good approach for your use case. @jglick any thoughts on whether exposing the ability to change the description of a job through `RunWrapper` seems ok in general?

jglick:
> any thoughts on whether exposing the ability to change the description of the job for the current build through `RunWrapper` is ok in general?

This does not feel right to me. I think this use case is better handled with `job-dsl`.

PR #101, opened 2020-01-10
setKeepLog from builds other than current

poinu:
We have a use case where we automatically mark some of the builds (releases) as "Keep forever". When a new build is marked, we would also like to be able to unmark the previous "Keep forever" builds, but the following check prevents us from doing so:

```java
public void setKeepLog(boolean b) throws IOException {
    if (!currentBuild) {
        throw new SecurityException("can only set the keepLog property on the current build");
    }

    //[...]
}
```

It would be nice if we could lift this restriction, or add a new method that allows this (and maybe requires the explicit approval from the Jenkins admin).


poinu:
Hi @dwnusbaum, thanks for your comment! Your suggestion regarding the global shared library does indeed make more sense. I implemented it that way and it is working, thanks for the tip :)

I'll close the PR.

PR #104, opened 2020-01-17
[JENKINS-55287] Add synchronization to BulkFlowNodeStorage

dwnusbaum:
See [JENKINS-55287](https://issues.jenkins-ci.org/browse/JENKINS-55287).

From a quick look at the issue, one hypothesis I had is that lack of synchronization causes the state in the Java code to not match the serialized state in some case, or for `isModified` to be `false` as seen by some thread that calls `flush` when it is really `true`, etc. If this does fix the issue, we'd obviously prefer to synchronize things at a more granular level, but synchronizing everything is much simpler and is obviously correct, so it should be fine for testing.

Even if this doesn't fix the issue, it's not clear to me why it is ok for these methods and similar ones in `SimpleXStreamFlowNodeStorage` to be unsynchronized.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
@@ -123,12 +123,12 @@ public class BulkFlowNodeStorage extends FlowNodeStorage {
 
     @Override
     @CheckForNull
-    public FlowNode getNode(@Nonnull String id) throws IOException {
+    public synchronized FlowNode getNode(@Nonnull String id) throws IOException {
         Tag t = getOrLoadNodes().get(id);
         return (t != null) ? t.node : null;
     }
 
-    public void storeNode(@Nonnull FlowNode n, boolean delayWritingActions) throws IOException {
+    public synchronized void storeNode(@Nonnull FlowNode n, boolean delayWritingActions) throws IOException {
         Tag t = getOrLoadNodes().get(n.getId());
         if (t != null) {
             t.node = n;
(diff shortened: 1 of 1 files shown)


DigitalTravelDuck:
I'm running constantly in this issue.
Error report (partly):
2020-01-25 09:24:01.881+0000 [id=20]    WARNING o.j.p.w.cps.CpsFlowExecution#onLoad: Error initializing storage and loading nodes, will try to create placeholders for: CpsFlowExecution[Owner[proj-Multibranch-Pipeline/ #buildnr]]
java.io.IOException: Tried to load head FlowNodes for execution Owner[proj-Multibranch-Pipeline/ #buildnr] but FlowNode was not found in storage for head id:FlowNodeId 1:63
        at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.initializeStorage(CpsFlowExecution.java:679)

And I'm willing to test.

dwnusbaum:
Closing and reopening to try to get an incremental build.

dwnusbaum:
Trying again, the Windows build failed due to infra issues: 

```
win2012-fa3830 was marked offline: Connection was broken: java.util.concurrent.TimeoutException: Ping started at 1580136351769 hasn't completed by 1580136591769
```

dwnusbaum:
@DigitalTravelDuck Thanks for the interest! Please try to add general comments about the problem in the Jira issue ([JENKINS-55287](https://issues.jenkins-ci.org/browse/JENKINS-55287)) instead of on GitHub. I added a comment to the issue with a link to the build if you want to try it out. It would be very helpful if you could upload a build folder of one of the builds that failed with that error when using the max survivability durability level as described in the description of the issue. This patch won't help you when using the max survivability durability level, and I haven't seen other users run into the issue in that case, so you might be experiencing a different issue.

DigitalTravelDuck:
Found the root cause, not enough memory on our test system, this killed
Jenkins.

Greetings,
Arnold.

dwnusbaum:
Various users reported that this didn't address the issue, so I am closing it for now.

PR #105, opened 2020-03-04
Use waitForMessage in tests to reduce flakes

timja:
Hi we're experiencing flakey tests trying to integrate the next bom line,

example:
```
Expected: a string containing "broken condition"
     but: was "Started
Running in Durability level: MAX_SURVIVABILITY
[Pipeline] Start of Pipeline
[Pipeline] waitUntil
[Pipeline] {
[Pipeline] semaphore
[Pipeline] }
Will try again after 0.25 sec
[Pipeline] {
[Pipeline] semaphore
[Pipeline] }
[Pipeline] // waitUntil
[Pipeline] End of Pipeline
```

@jglick suggested replacing assertLogContains with waitForMessage
https://github.com/jenkinsci/bom/pull/110#issuecomment-594782396

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/SerializationSecurityTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/SerializationSecurityTest.java
@@ -256,8 +256,8 @@ public class SerializationSecurityTest {
     private static void safe(JenkinsRule r, WorkflowRun b) throws Exception {
         assertNull(r.jenkins.getSystemMessage());
         assertEquals(Result.FAILURE, b.getResult());
+        r.waitForMessage("staticMethod jenkins.model.Jenkins getInstance", b);
         r.assertLogNotContains("should not still have", b);
-        r.assertLogContains("staticMethod jenkins.model.Jenkins getInstance", b);
     }
 
 }


dwnusbaum:
I feel like it would make more sense to fix workflow-job so that `assertLogContains` isn't flaky (there are [possible ideas](https://github.com/jenkinsci/workflow-job-plugin/pull/131#discussion_r291657569) in this thread) rather than just changing tests everywhere.

timja:
> I feel like it would make more sense to fix workflow-job so that `assertLogContains` isn't flaky when used after the build is supposed to have completed (there are [possible ideas](https://github.com/jenkinsci/workflow-job-plugin/pull/131#discussion_r291657569) in this thread) rather than just changing tests everywhere.

Having a look, but is there any harm in making the tests more resilient with this change?

currently if I ran a build random tests in 1 or 2 out of a list of ~5 plugins will always fail 1-2 tests

jglick:
> make more sense to fix workflow-job

Ultimately yes, but this carries a relatvely high risk of regression I think.

src/test/java/org/jenkinsci/plugins/workflow/support/actions/LogActionImplTest.java
```diff
@@ -54,12 +54,12 @@
...
-        r.assertLogContains("atom step in A with 1 commands to go", b);
-        r.assertLogContains("logging from LBBL with 0 commands to go", b);
-        r.assertLogContains("atom step in B with 1 commands to go", b);
-        r.assertLogContains("logging from BL with 0 commands to go", b);
+        r.waitForMessage("logging from LBBL with 3 commands to go", b);
+        r.waitForMessage("atom step in A with 2 commands to go", b);
+        r.waitForMessage("atom step in A with 1 commands to go", b);
+        r.waitForMessage("logging from LBBL with 0 commands to go", b);
+        r.waitForMessage("atom step in B with 1 commands to go", b);
+        r.waitForMessage("logging from BL with 0 commands to go", b);
```
jglick: Replacing all occurrences is probably overkill. AFAIK the issue occurs when a message is printed right when the build is ending, such as the final error/stack trace in `WorkflowRun.finish`. So would suffice to patch just the cases observed to fail.

timja:
Closing stale PRs of mine

jglick:
IIRC this would now be fixed just by using a sufficiently new `jenkins-test-harness`.

PR #108, opened 2020-05-27
[JENKINS-62545] Attempt to detect cycles when iterating through siblings in FlowGraphTable

dwnusbaum:
See [JENKINS-62545](https://issues.jenkins-ci.org/browse/JENKINS-62545).

I saw a report recently of a thread stuck inside of the `while` loop in `FlowGraphTable.Row.addTreeSibling`. I am not sure how this could be happening, and I do not have any reproduction steps, but the only thing that comes to mind is an infinite loop caused by a cycle in the graph (probably not the actual flow graph, just in the representation in `FlowGraphTable`).

Here is a draft PR that tries to detect cycles so that we can understand what the real bug is. I have not tested it yet, and I am not sure if this code is covered by unit tests or not, so it needs to be verified manually.

CC @res0nance 

EDIT: I got more info, so now I know the proximate cause of the issue, although I am still not sure about the root cause. It turns out that the actual flow graph itself was corrupted. Here is a simplified form of its structure, the `sh` step is the problematic node:

```
               b1Start-...-b1End
              /                 \
start-parStart-b2Start-...-b2End-parEnd-...-end
              \                 /          /
               b3Start-...----sh--...-b3End
```
The problem is that the `sh` step is the parent of the end of a parallel step, but somehow also the parent of additional steps. `FlowGraphTable` doesn't expect the graph to have a structure like this so things break. The build was manually aborted while the parallel branches were executing, yet somehow branch 3 continued executing despite the end node for the parallel step being linked to one of the nodes in that branch. I don't have access to Jenkins system logs from when the build was executed to know if there were any warnings. The Pipeline in question looks relatively normal (there aren't nested parallels or anything like that).

I am not sure how this could happen, and I wasn't able to reproduce the issue from scratch.

I think it's reasonable to make `FlowGraphTable` throw an exception for a corrupted graph like this to avoid infinite loops.


res0nance:
Hey Devin 👋 , thanks for looking into this. I've a build that thankfully still exhibits this issue. Since it might contain sensitive data, I'll reach out to you out of band on this.

dwnusbaum:
CC @jglick have you ever seen a flow graph with the structure in the description of this PR? Any idea of what could cause it?

jglick:
No idea what the root cause could be, and I agree that throwing something like an `IllegalStateException` in these circumstances would be appropriate.

dwnusbaum:
There was no test coverage for `FlowGraphTable` in this plugin, so I added some smoke tests and a regression test for JENKINS-62545.

PR #110, opened 2020-11-05
[JENKINS-47101] Add a default EnvironmentExpander that registers password parameters as sensitive variables

dwnusbaum:
See [JENKINS-47101](https://issues.jenkins-ci.org/browse/JENKINS-47101).

Before the fix for JENKINS-47101 in https://github.com/jenkinsci/workflow-cps-plugin/pull/370, password parameters would have been masked in step arguments, so I think we need to make sure that they continue to be masked, which requires us to integrate with the new API from https://github.com/jenkinsci/workflow-step-api-plugin/pull/57.

We need to add a test for this case over in `workflow-cps`.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
@@ -31,11 +31,18 @@ import hudson.console.ConsoleLogFilter;
 import hudson.model.Computer;
 import hudson.model.Job;
 import hudson.model.Node;
+import hudson.model.ParameterValue;
+import hudson.model.ParametersAction;
+import hudson.model.PasswordParameterValue;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import java.io.IOException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -97,6 +104,9 @@
             return castOrNull(key,getExecution());
         } else if (FlowNode.class.isAssignableFrom(key)) {
             return castOrNull(key, getNode());
+        } else if (key == EnvironmentExpander.class) {
+            // Only called when `value == null` so that it does not override expanders contributed by steps.
+            return key.cast(new PasswordParameterEnvironmentExpander(get(Run.class)));
```
dwnusbaum, line 109: For now we just have single default implementation since I cannot think of any other similar cases to password parameters. If there are similar cases, maybe we should create an extension point for marking sensitive variables that are active from the start of the build or something.

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -97,6 +104,9 @@
             return castOrNull(key,getExecution());
         } else if (FlowNode.class.isAssignableFrom(key)) {
             return castOrNull(key, getNode());
+        } else if (key == EnvironmentExpander.class) {
```
dwnusbaum, line 107: Maybe I should move this case up next to the `TaskListener` case?

car-roll (approved):
LGTM

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
car-roll: Do you think the password parameter case may not get reached?

dwnusbaum: No, just for consistency/style, since the cases here are checking `isAssignableFrom`, I think in case someone looks up a subclass of the type, but for this case, I don't think that would make sense, and the direct-equality cases are handled above.

dwnusbaum: Moved up in b142593 (I force-pushed over the old commit)

PR #112, opened 2020-11-25
[JENKINS-64282] Do not add empty passwords to sensitive variables list

car-roll:
Prevents empty value password parameters from being added to the sensitive variables list in `workflow-api`

https://github.com/jenkinsci/workflow-cps-plugin/pull/401 does the same thing, but that is more of a general hardening of `workflow-cps`

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
@@ -195,7 +195,7 @@ public abstract class DefaultStepContext extends StepContext {
             ParametersAction action = run.getAction(ParametersAction.class);
             if (action != null) {
                 passwordParameterVariables = action.getParameters().stream()
-                        .filter(e -> PasswordParameterValue.class.isInstance(e)
+                        .filter(e -> e instanceof PasswordParameterValue
                                         && !((Secret) e.getValue()).getPlainText().isEmpty())
                         .map(ParameterValue::getName)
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -194,7 +195,8 @@ public PasswordParameterEnvironmentExpander(Run<?, ?> run) {
             ParametersAction action = run.getAction(ParametersAction.class);
             if (action != null) {
                 passwordParameterVariables = action.getParameters().stream()
-                        .filter(PasswordParameterValue.class::isInstance)
+                        .filter(e -> PasswordParameterValue.class.isInstance(e)
```
dwnusbaum, line 198: nit: I was using `Class.isInstance` to avoid a lambda, but since we have the lambda now anyway it might be clearer to write `e instanceof PasswordParameterValue` instead.

src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
```diff
@@ -194,7 +195,8 @@ public PasswordParameterEnvironmentExpander(Run<?, ?> run) {
...
+                                        && !((Secret) e.getValue()).getPlainText().isEmpty())
```
dwnusbaum, line 199: I was worried about possible NPEs here but it looks like `PasswordParameterValue.getValue` and `Secret.getPlainText` are both annotated with `@NonNull` so it should be fine.

dwnusbaum (approved):

PR #114, opened 2021-01-16
Make workflow-support compatible with Guava 21.0 and newer

timja:
See https://github.com/jenkinsci/jenkins/pull/5059.

`workflow-support` previously used `MoreExecutors.sameThreadExecutor` from Guava, which was [removed in version 21.0](https://github.com/google/guava/wiki/Release21#commonutilconcurrent) in favor of `MoreExecutors.newDirectExecutorService`. This PR uses reflection to first try to call `MoreExecutors.newDirectExecutorService` but then call `MoreExecutors.sameThreadExecutor` if the first method is not available to support both older and newer versions of Guava.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-support</artifactId>
-    <version>3.7</version>
+    <version>${revision}${changelist}</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Supporting APIs</name>
     <url>https://github.com/jenkinsci/workflow-support-plugin</url>
@@ -47,7 +47,7 @@
         <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
         <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
         <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
-        <tag>workflow-support-3.7</tag>
+        <tag>${scmTag}</tag>
     </scm>
     <repositories>
         <repository>
@@ -62,7 +62,7 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <revision>3.7</revision>
+        <revision>3.8</revision>
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.176.4</jenkins.version>
         <java.level>8</java.level>
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-support</artifactId>
-    <version>${revision}${changelist}</version>
+    <version>3.7</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Supporting APIs</name>
     <url>https://github.com/jenkinsci/workflow-support-plugin</url>
@@ -47,7 +47,7 @@
         <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
         <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
         <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
-        <tag>${scmTag}</tag>
+        <tag>workflow-support-3.7</tag>
     </scm>
     <repositories>
         <repository>
--- a/pom.xml
+++ b/pom.xml
@@ -44,9 +44,9 @@
         </license>
     </licenses>
     <scm>
-        <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
-        <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
-        <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
+        <connection>scm:git:git://github.com/${gitHubRepo}.git</connection>
+        <developerConnection>scm:git:git@github.com:${gitHubRepo}.git</developerConnection>
+        <url>https://github.com/${gitHubRepo}</url>
         <tag>${scmTag}</tag>
     </scm>
     <repositories>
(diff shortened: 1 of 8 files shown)


timja:
seriously -.-, old pinned version in Jenkinsfile...

pom.xml
```diff
@@ -64,11 +64,12 @@
     <properties>
         <revision>3.8</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.222.4</jenkins.version>
+        <jenkins.version>2.273-rc30689.09147672b85b</jenkins.version>
```
dwnusbaum, line 67: Can you please link to whatever Jenkins core PR produced this incremental so reviewers have more context about the proposed change? Thanks!

timja, line 67:
```suggestion
        <jenkins.version>2.273-rc30689.09147672b85b</jenkins.version> <!-- https://github.com/jenkinsci/jenkins/pull/5059 -->
```

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Futures.java
```diff
@@ -324,7 +324,7 @@ public void run() {
     public static <V> ListenableFuture<List<V>> allAsList(
         ListenableFuture<? extends V>... futures) {
       return new ListFuture<V>(ImmutableList.copyOf(futures), true,
-          MoreExecutors.sameThreadExecutor());
+          MoreExecutors.newDirectExecutorService());
```
jglick, line 327: Since this is since Guava 18, maybe we could use reflection to try `newDirectExecutorService` then fall back to `sameThreadExecutor`, then revert the core bump?

Or copy the impl into this plugin, keeping the Apache license header.

timja: nice idea, when I was working on it (and plan to get back to it, although others are welcome to take it up) I was trying to get a green bom PCT build and then look at what's required and see where changes like your suggestion could be applied.

amuniz: I have this change ready to push. Is it ok if I push to your branch @timja?

timja: absolutely go for it

timja: also see https://github.com/jenkinsci/workflow-api-plugin/pull/135 and https://github.com/jenkinsci/bom/pull/423

amuniz: Errr, I don't have push permission here, so I've filed a PR to your forked branch: https://github.com/timja/workflow-support-plugin/pull/1

amuniz: > also see jenkinsci/workflow-api-plugin#135 and jenkinsci/bom#423

I'll have a look. Meanwhile, isn't this PR mergeable as is?

pom.xml
```diff
@@ -97,6 +98,12 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>script-security</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.checkerframework</groupId>
+                    <artifactId>checker-qual</artifactId>
+                </exclusion>
+            </exclusions>
```
timja, line 106:
```suggestion
```

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Futures.java
timja: looks good to me

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Futures.java
```diff
@@ -304,6 +307,28 @@ public void run() {
       return chain(future, wrapperFunction, executor);
     }
 
+    /**
+     * Returns an {@link ExecutorService} to be used as a parameter in other methods.
+     * It calls {@code MoreExecutors#newDirectExecutorService} or falls back to {@code MoreExecutors#sameThreadExecutor}
+     * for compatibility with older (< 18.0) versions of guava.
```
jglick, line 313:
```suggestion
     * for compatibility with older (&lt; 18.0) versions of guava.
```

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Futures.java
```diff
@@ -304,6 +307,28 @@ public void run() {
...
+     *
+     * @since TODO
+     */
+    public static ExecutorService newExecutorService() {
+        try {
+            try {
+                Method method = MoreExecutors.class.getMethod("newDirectExecutorService");
+                return (ExecutorService) method.invoke(null);
+            } catch (NoSuchMethodException e) {
+                // guava older than 18, fallback to `sameThreadExecutor`
```
jglick, line 323:
```suggestion
                // Guava older than 18, fall back to `sameThreadExecutor`
```

jglick (approved):

dwnusbaum (approved):

dwnusbaum:
I updated the PR title and description to reflect the current proposed change.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Futures.java
```diff
@@ -304,6 +307,28 @@ public void run() {
...
+     */
+    public static ExecutorService newExecutorService() {
+        try {
+            try {
+                Method method = MoreExecutors.class.getMethod("newDirectExecutorService");
+                return (ExecutorService) method.invoke(null);
+            } catch (NoSuchMethodException e) {
+                // Guava older than 18, fall back to `sameThreadExecutor`
+                Method method = MoreExecutors.class.getMethod("sameThreadExecutor");
+                return (ExecutorService) method.invoke(null);
```
dwnusbaum, line 325: nit: Maybe it would be better to reverse the order until Guava is updated in Jenkins core to avoid throwing `NoSuchMethodException` on a regular basis.

car-roll (approved):

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Futures.java
timja: done

car-roll (approved):

jglick:
@car-roll this will need a release too. (Or https://github.com/jenkinsci/incrementals-tools#automatic-deployment if you are sick of my bugging you about stuff like this :grin:.)

car-roll:
> if you are sick of my bugging you about stuff like this 
Not yet! 😁  But I should probably look into that stuff still...
Will release later today

PR #117, opened 2021-03-14
Vendored Guava should not use reflection

basil:
Simpler and reflection-free alternative to #114 (CC @timja).

`org.jenkinsci.plugins.workflow.support.concurrent.ChainingListenableFuture`, `org.jenkinsci.plugins.workflow.support.concurrent.Futures`, and `org.jenkinsci.plugins.workflow.support.concurrent.ListFuture` are private copies of Guava classes specifically for use by Pipeline (in particular, they are being used by `workflow-cps`, `workflow-cps-checkpoint`, `workflow-job`, and `workflow-support`). The reasoning for making private copies is given in Kohsuke's comment:

> Mostly copied after Guava's `Futures`, because that one is still marked as beta and is subject to change.

The whole point of vendoring these classes was to insulate Pipeline from Guava API changes. The problem here is that Kohsuke just didn't go far enough with this design, since what he copied and pasted itself depended on APIs marked as beta and subject to change. This change just completes what Kohsuke started by vendoring the rest of the Guava related code in the call hierarchy. I also cherry-picked some fixes and updates from upstream to get the vendored version up-to-date.

This is all consistent with the original design of vendoring the relevant `Futures` functionality for use in Pipeline. But one might also ask if this design still makes sense in a Guava 30+ world where these APIs are no longer beta. Of course it doesn't, so once Jenkins core is on a newer version of Guava and the baseline has been updated in all the Pipeline plugins, these vendored classes can be removed. Accordingly I've marked the existing ones as `@Deprecated` and the new ones as `@Restricted(DoNotUse.class)` in anticipation of this eventual removal.

I'll test this with `workflow-cps` and `workflow-job` which use the vendored classes. It would be great if a CloudBees employee could test this change with `workflow-cps-checkpoint` which I don't have access to.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/ChainingListenableFuture.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/ChainingListenableFuture.java
@@ -3,7 +3,6 @@ package org.jenkinsci.plugins.workflow.support.concurrent;
 import com.google.common.util.concurrent.AbstractFuture;
 import com.google.common.util.concurrent.AsyncFunction;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.MoreExecutors;
 
 import javax.annotation.Nullable;
 import java.lang.reflect.UndeclaredThrowableException;
@@ -208,7 +207,7 @@ class ChainingListenableFuture<I, O>
               ChainingListenableFuture.this.outputFuture = null;
             }
           }
-        }, Futures.newExecutorService());
+        }, MoreExecutors.directExecutor());
     } catch (UndeclaredThrowableException e) {
       // Set the cause of the exception as this future's exception
       setException(e.getCause());
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/ChainingListenableFuture.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/ChainingListenableFuture.java
@@ -29,6 +29,7 @@ import static java.util.concurrent.TimeUnit.NANOSECONDS;
  * in a {@code UndeclaredThrowableException} so that this class can get
  * access to the cause.
  */
+@Deprecated
 class ChainingListenableFuture<I, O>
     extends AbstractFuture<O> implements Runnable {
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/DirectExecutor.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/DirectExecutor.java
@@ -15,6 +15,8 @@
 package org.jenkinsci.plugins.workflow.support.concurrent;
 
 import com.google.common.annotations.GwtCompatible;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
 import java.util.concurrent.Executor;
 
 /**
@@ -22,6 +24,7 @@ import java.util.concurrent.Executor;
  * execute}.
  */
 @GwtCompatible
+@Restricted(NoExternalUse.class)
 enum DirectExecutor implements Executor {
   INSTANCE;
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/DirectExecutor.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/DirectExecutor.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2007 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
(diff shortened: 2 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/ChainingListenableFuture.java
```diff
@@ -30,6 +29,7 @@
  * in a {@code UndeclaredThrowableException} so that this class can get
  * access to the cause.
  */
+@Deprecated
```
basil, line 32: This class is only used from `workflow-support`. Deprecating it so that `workflow-support` can prefer the upstream version once core is on a recent version of Guava and `workflow-support`'s baseline has been updated.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/ChainingListenableFuture.java
```diff
@@ -208,7 +208,7 @@ public void run() {
               ChainingListenableFuture.this.outputFuture = null;
             }
           }
-        }, Futures.newExecutorService());
+        }, MoreExecutors.directExecutor());
```
basil, line 211: Undoing a change made in #114 in favor of the upstream version. We should generally avoid changing vendored upstream code.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/DirectExecutor.java
```diff
@@ -0,0 +1,40 @@
...
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import java.util.concurrent.Executor;
+
+/**
+ * An {@link Executor} that runs each task in the thread that invokes {@link Executor#execute
+ * execute}.
+ */
+@GwtCompatible
+@Restricted(NoExternalUse.class)
```
basil, line 27: Taken without change from Guava 30.1. Used only by the vendored code. Marked as restricted to prevent any further usages.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Futures.java
```diff
@@ -45,6 +41,7 @@
  *
  * @author Guava
  */
+@Deprecated
```
basil, line 44: This class is used from `workflow-cps`, `workflow-cps-checkpoint`, `workflow-job`, and `workflow-support`. Deprecating it so that these plugins can prefer the upstream version once core is on a recent version of Guava and the plugins' baselines have been updated.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/MoreExecutors.java
```diff
@@ -0,0 +1,88 @@
...
+ * Factory and utility methods for {@link java.util.concurrent.Executor}, {@link ExecutorService},
+ * and {@link java.util.concurrent.ThreadFactory}.
+ *
+ * @author Eric Fellheimer
+ * @author Kyle Littlefield
+ * @author Justin Mahoney
+ * @since 3.0
+ */
+@GwtCompatible(emulated = true)
+@Restricted(NoExternalUse.class)
```
basil, line 35: Taken without change from Guava 30.1. Used only by the vendored code. Marked as restricted to prevent any further usages.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/ListFuture.java
```diff
@@ -37,6 +36,7 @@
    * each component future to fill out the value in the List when that future
    * completes.
    */
+@Deprecated
```
basil, line 39: This class is only used from workflow-support. Deprecating it so that workflow-support can prefer the upstream version once core is on a recent version of Guava and workflow-support's baseline has been updated.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/DirectExecutor.java
timja: should this be added as a comment in the source?

basil: I think the commit message should cover the provenance well enough:

```
commit ec98115 Add MoreExecutors#directExecutor() implementation from Guava 30.1
```

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/Futures.java
jglick: Which could now be whenever we like, right?

Also some of these methods could be deprecated in favor of Java Platform equivalents, like `CompletableFuture`.

car-roll: Are we okay with doing that in a separate PR?

jglick: Sure.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/DirectExecutor.java
jglick: Though `git blame` will not show that if this PR is squash-merged, as @car-roll normally seems to do.

car-roll: yeah, i noticed that line too. I was going to make a special exception :)

car-roll (approved):

PR #120, opened 2021-06-30
Improve durability of `SURVIVABLE_NONATOMIC` while retaining the skipping of `fsync(2)`

basil:
Not yet ready for review. Filing just to get an incremental build.

See [JENKINS-66001](https://issues.jenkins.io/browse/JENKINS-66001). Downstream of jenkinsci/jenkins#5599. Relates to jenkinsci/workflow-api-plugin#163. Part 3 of a 5-part series to make Pipeline's `SURVIVABLE_NONATOMIC` mode behave [as advertised in the documentation](https://www.jenkins.io/doc/book/pipeline/scaling-pipeline/#what-am-i-giving-up-with-this-durability-setting-trade-off) (parts 4 through 5 will be in `workflow-cps` and `workflow-job` respectively).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -23,12 +23,12 @@
   ~ THE SOFTWARE.
   -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.15</version>
+        <version>4.19</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -64,7 +64,7 @@
     <properties>
         <revision>3.9</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.222.4</jenkins.version>
+        <jenkins.version>2.300-rc31329.23a941f5b44f</jenkins.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/5599 -->
         <java.level>8</java.level>
         <no-test-jar>false</no-test-jar>
         <useBeta>true</useBeta>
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/PipelineIOUtils.java
```diff
@@ -23,11 +25,36 @@
...
+            @Nonnull Object toWrite,
+            @Nonnull File location,
+            @Nonnull XStream2 xstream,
+            boolean atomicWrite,
+            boolean force)
+            throws IOException {
         if (atomicWrite) {
             XmlFile file = new XmlFile(xstream, location);
-            file.write(toWrite);
+            file.write(toWrite, force);
```
bitwiseman, line 57: Is there any way to do this such that the plugin doesn't have to be limited to Jenkins versions with the upstream change?  If not, we'll have to maintain split release trains, which isn't terrible but is added overhead.

basil: Not that I'm aware of, I'm afraid.

bitwiseman: @basil 
Excuse me, what I mean is using reflection such as in https://github.com/jenkinsci/scm-api-plugin/commit/783b9824f567ec1cbccc91f601064a932af59636 

Or would that result in horrible performance?

basil: Not sure about performance, but using reflection to avoid core API bumps has historically caused a number of problems in Pipeline, including jenkinsci/workflow-basic-steps-plugin#127. In general I think Jesse's preference has historically been to keep things simple and avoid reflection for that reason.

pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>3.9</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.222.4</jenkins.version>
+        <jenkins.version>2.300-rc31329.23a941f5b44f</jenkins.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/5599 -->
```
jglick, line 67: I recommend marking downstream PRs like this as draft, to make sure they are not accidentally merged prematurely.

src/main/java/org/jenkinsci/plugins/workflow/support/PipelineIOUtils.java
jglick: Yup. If and when you actually come across a bug fix that you wish to make available to users of LTS or even older versions, https://gist.github.com/jglick/86a30894446ed38f918050c1180483e2 is easy enough. Not worth making life complicated “just in case” that happens.

jtnord:
@jglick My comments on the upstream are not addressed, and this can not be reviewed in isolation.

jtnord (commented):
My comments on the upstream change are not addressed, and this can not be reviewed in isolation.

basil:
Your comments are not and will not be addressed.

PR #123, opened 2021-11-16
Attempting to diagnose opaque error from `RiverReader.readPickles`

jglick:
See https://github.com/jenkinsci/workflow-cps-plugin/pull/480 for background.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverReader.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverReader.java
@@ -170,6 +170,9 @@ public class RiverReader implements Closeable {
 
     @SuppressWarnings("unchecked")
     private List<Pickle> readPickles(int offset) throws IOException {
+        if (offset == 0) {
+            throw new IOException("No offset for pickles");
+        }
         try (BufferedInputStream es = openStreamAt(offset)) {
             MarshallingConfiguration config = new MarshallingConfiguration();
             config.setClassResolver(new SimpleClassResolver(classLoader));
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverReader.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverReader.java
@@ -118,7 +118,7 @@ public class RiverReader implements Closeable {
             throw new IOException("Invalid stream header");
 
         short v = din.readShort();
-        if (v!=1)
+        if (v!= RiverWriter.VERSION)
             throw new IOException("Unexpected stream version: "+v);
 
         return din.readInt();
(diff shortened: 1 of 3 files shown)


dwnusbaum (approved):

jglick:
Still trying to ascertain whether https://github.com/jenkinsci/workflow-support-plugin/blob/fe97851d67bcfe9e21bf108c4d7a888ad9892c25/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java#L175-L182 is ever called, and if so, why `program.dat` still has 0x00000000 at offset 10.

jglick:
This is happening in the context of AWS EBS snapshots, which operate on 512Kib blocks. Some of the `program.dat` files exceed that size. I suspect the problem is that despite the supposedly atomic move of a temp file, `program.dat` is actually being observed by the snapshot process in a half-written state. See also #120 from @basil dealing with `workflow/*.xml` files.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverReader.java
```diff
@@ -164,9 +168,12 @@ public Unmarshaller apply(PickleResolver input) {
         }
     }
 
+    @SuppressWarnings("unchecked")
     private List<Pickle> readPickles(int offset) throws IOException {
-        BufferedInputStream es = openStreamAt(offset);
-        try {
+        if (offset == 0) {
+            throw new IOException("No offset for pickles");
```
jglick, line 174: Observed this error in a controller I am testing, so the code in the patch works as expected. Unfortunately I was not able to locate a log file covering the time period of the particular build with the error so I am not sure yet whether the fine logging added to `RiverWriter` will be helpful in tracking down the root cause; I do see these three messages being printed in the expected order for other builds.

jglick (commented):
@car-roll if this looks OK, could it be merged please? It should it least make certain errors easier to diagnose than the previously totally opaque message. I have other changes to this code area I am considering and it would be easiest to start from this PR merged into trunk.

dwnusbaum:
> program.dat is actually being observed by the snapshot process in a half-written state

So you mean the atomic move is not atomic at the underlying storage level? Or something else? Would using `FileChannel` instead of `RandomAccessFile` to write the pickle offset and calling `force(true)` make a difference?

Also, would it be possible or useful to write some kind of sentinel value between the main program and the pickles, e.g. `0xABCDEFABCDEFABCEDFABCDEF`, so that we can tell just from looking at the hex dump printed by `RiverReader` whether the sentinel value and/or the pickles are present?

car-roll (approved):

jglick:
> So you mean the atomic move is not atomic at the underlying storage level?

That is my guess. The FS here is ext4. Perhaps that does not guarantee that a random-access file write followed by an inode move is ordered?

> Would using `FileChannel` instead of `RandomAccessFile` to write the pickle offset and calling `force(true)` make a difference?

Maybe? I am no expert in this topic. I can try that and see if the error disappears, though it was intermittent to begin with so it is hard to prove.

One thing I have noticed is that `program.dat` can contain a lot of redundancy in text strings. I was considering gzipping it but now I am thinking it may be better to just write a ZIP file, with one (compressed!) entry for the program and one for the pickles, which would probably be simpler than the current trick. Not clear it would prevent filesystem-level errors, though.

dwnusbaum:
I filed [JENKINS-67389](https://issues.jenkins.io/browse/JENKINS-67389) to cover this topic and the related PRs.

PR #132, opened 2021-12-16
Replace Guava cache with Caffeine

Vlatombe:
* Replace guava cache with Caffeine

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
@@ -71,8 +71,7 @@ public class SimpleXStreamFlowNodeStorage extends FlowNodeStorage {
     private final FlowExecution exec;
 
     private final LoadingCache<String,FlowNode> nodeCache = Caffeine.newBuilder()
-            .weakKeys()
-            .weakValues()
+            .softValues()
             .build(key -> SimpleXStreamFlowNodeStorage.this.load(key).node);
 
     private static final Logger LOGGER = Logger.getLogger(SimpleXStreamFlowNodeStorage.class.getName());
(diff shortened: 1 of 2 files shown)


jglick:
Relates to https://github.com/jenkinsci/workflow-cps-plugin/pull/492.

jglick (commented):
Caffeine switch is certainly good. I am not convinced we want to switch from soft to weak references; seems like it would wind up causing increased disk I/O (and some CPU) under typical load. The root cause of the large heap allocations and long GC pauses mentioned in the description turned out to be incomplete truncation behavior in `ArgumentsActionImpl`, which is already being corrected in a separate PR, so this PR would penalize instances with more normal metadata (and just a lot of builds and steps).

src/main/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamFlowNodeStorage.java
```diff
@@ -71,11 +69,11 @@
...
     private final File dir;
     private final FlowExecution exec;
-    private final LoadingCache<String,FlowNode> nodeCache = CacheBuilder.newBuilder().softValues().build(new CacheLoader<String,FlowNode>() {
-        @Override public FlowNode load(String key) throws Exception {
-            return SimpleXStreamFlowNodeStorage.this.load(key).node;
-        }
-    });
+
+    private final LoadingCache<String,FlowNode> nodeCache = Caffeine.newBuilder()
+            .weakKeys()
```
jglick, line 74: Also, why weak _keys_? These should be small `String`s, currently always numeric like `123`. We could perhaps `intern` them, but beyond that…?

Vlatombe:
I agree with you, the cache problem here is a consequence of the incomplete truncation. Better progress on the other, and possibly keep the switch to Caffeine. Soft vs. weak references could be reconsidered later.

car-roll:
LGTM  if we just do the switch to caffeine

jglick (approved):

car-roll (approved):

PR #129, opened 2021-12-14
[JENKINS-67389] Use FileChannel to save program.dat to avoid corruption

dwnusbaum:
See [JENKINS-67389](https://issues.jenkins.io/browse/JENKINS-67389).

Related to #123 and https://github.com/jenkinsci/workflow-cps-plugin/pull/480. From some quick testing this seems to help with the corruption issues that @jglick and I have seen when backing up and restoring Jenkins while it is running.

Before this PR, we would open one `FileOutputStream` to write the majority of `program.dat` (to a temp file here), then close that stream and open a `RandomAccessFile` just to write the pickle offset. We never invoked `FileChannel.force` (i.e. `fsync`), and so I think the filesystems we have been testing were happy to rename the temp file to `program.dat` before the writes to the file were fully committed to disk, leading to a case where the pickle offset was not yet written (Jesse saw this) and a case where `program.dat` existed but was 0 bytes (I saw this).

With this PR, we open a single `FileChannel` to write all of `program.dat`, and we invoke `FileChannel.force` in an attempt to prevent the filesystem from reordering the write and the rename so that `program.dat` never exists in a partially-written state.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #151, opened 2022-03-04
Buffer writes on the Java side in RiverWriter

dwnusbaum:
I saw a stack trace indicating that #129 may have introduced performance issues in some cases due to lack of write buffering on the Java side. `FileChannel.write` directly calls OS APIs, so while there is still potentially an OS-side buffer, it might be slower than buffering writes on the Java side.

Untested, and I have not attempted to validate the performance impact.

- [x] Link to relevant issues in GitHub or Jira

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java
@@ -188,8 +188,7 @@ public class RiverWriter implements Closeable {
 
     private static class FileChannelOutput implements ByteOutput {
         private final FileChannel channel;
-        /** Used to reduce allocation for single-byte writes. */
-        private final byte[] singleton = new byte[1];
+        private final ByteBuffer buffer = ByteBuffer.allocate(8192); // c.f. BufferedOutputStream
 
         private FileChannelOutput(FileChannel channel) {
             this.channel = channel;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java
```diff
@@ -188,36 +188,53 @@ public void close() throws IOException {
...
+            }
+            if (len > buffer.remaining()) {
+                writeBuffer();
+            }
+            buffer.put(bytes, off, len);
+        }
+
+        private void writeBuffer() throws IOException {
+            buffer.flip();
+            channel.write(buffer);
```
dwnusbaum, line 225: Better to not even call `channel.write` if the buffer is empty. Might happen if `flush` and `close` get called back-to-back in some case.

```suggestion
            if (buffer.limit() != 0) {
                channel.write(buffer);
            }
```

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java
```diff
@@ -188,36 +188,55 @@ public void close() throws IOException {
...
+                channel.write(ByteBuffer.wrap(bytes, off, len));
+                return;
+            }
+            if (len > buffer.remaining()) {
+                writeBuffer();
+            }
+            buffer.put(bytes, off, len);
+        }
+
+        private void writeBuffer() throws IOException {
```
jglick, line 223: Or it might be clearer to just make this be `flush()`.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/RiverWriter.java
```diff
@@ -188,36 +188,55 @@ public void close() throws IOException {
...
+            buffer.flip();
+            if (buffer.limit() > 0) {
+                channel.write(buffer);
```
jglick, line 226: Do we need to check the return value? (The code did not do so before. Was unclear to me from Javadoc whether `FileChannel` is guaranteed to write all available bytes.)

jglick (approved):
Looks reasonable. Might be useful to have an `@Ignore`d addition to `RiverWriterTest` which writes something big—maybe a large array of large arrays—and displays elapsed `nanoTime`, before and after this patch.

dwnusbaum:
> Might be useful to have an @Ignored addition to RiverWriterTest which writes something big—maybe a large array of large arrays—and displays elapsed nanoTime, before and after this patch.

I made a quick test like that, and the performance was essentially the same between both versions of the code on my machine.

Turns out that `RiverMarshaller` extends `AbstractMarshaller` extends `AbstractObjectOutput` extends [`SimpleDataOutput`](https://github.com/jboss-remoting/jboss-marshalling/blob/9967cb7d4fcdebca515dc8ead4e53cd1161f4910/api/src/main/java/org/jboss/marshalling/SimpleDataOutput.java#L29), and `SimpleDataOutput` has its own buffer (8192 bytes by default) in front of our custom `FileChannelOutput` (I have a vague memory now of looking into this back when I created #129), so as far as I can tell, adding an additional buffer to `FileChannelOutput` should not really make any difference in terms of performance.

Based on that, I'm going to close this PR unless anyone has any objections.

PR #157, opened 2022-03-29
Remove unnecessary use of reflection

basil:


basil:
Lights are on, but apparently nobody is home.

PR #161, opened 2022-04-06
Don't show workspace icon if the user lacks permission to access it

NotMyFault:
The change proposed hides the `Workspaces` button if the user lacks permission to access it.

The current behavior of the icon being shown can be seen on ci.jenkins.io, for example: https://ci.jenkins.io/job/Core/job/jenkins/job/master/lastSuccessfulBuild/flowGraphTable/

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceRunAction.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceRunAction.java
@@ -59,19 +59,20 @@ public final class WorkspaceRunAction implements Action {
         this.owner = owner;
     }
 
+    private boolean hasNoWorkspacePermission() {
+        return (build instanceof AccessControlled && !((AccessControlled) build).hasPermission(Item.WORKSPACE));
+    }
+
     @Override public String getIconFileName() {
-        return "folder.png";
+        return hasNoWorkspacePermission() ? null : "folder.png";
     }
 
     @Override public String getDisplayName() {
-        return Messages.workspaces();
+        return hasNoWorkspacePermission() ? null : Messages.workspaces();
     }
 
     @Override public String getUrlName() {
-        if (build instanceof AccessControlled && !((AccessControlled) build).hasPermission(Item.WORKSPACE)) {
-            return null;
-        }
-        return "ws";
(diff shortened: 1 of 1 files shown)


NotMyFault:
Heyo, do you mind taking a look @car-roll? :)

src/main/java/org/jenkinsci/plugins/workflow/support/actions/WorkspaceRunAction.java
```diff
@@ -59,19 +59,20 @@ public final class WorkspaceRunAction implements Action {
         this.owner = owner;
     }
 
+    private boolean canAccess() {
+        return (build instanceof AccessControlled && !((AccessControlled) build).hasPermission(Item.WORKSPACE));
```
car-roll, line 63: This might be a nit, but when I think of `canAccess`, I would think that a `true` would mean that I do have access. In this case, `canAccess` is returning `false`, it feels a bit confusing for me.

NotMyFault: I've renamed the method.

car-roll: Sorry, I think I am confusing myself. Please correct me if I am wrong here:

In the code here, the method is returning `true` when there is no workspace permission. But wouldn't the name `hasWorkspacePermission` or `canAccess` mean that when it returns `true` the user has permission?

Unless I am mistaken and that you actually do return `null` when the user has workspace permission?

NotMyFault: It's the same condition used before to check whether the user has access to the workspace, just wrapped into a private method to reuse a bunch of times.

The code does return false if the user lacks permission to access the workspace or build is no instance of AccessControlled, and in this case the icon to access the workspace is hidden from the UI.

car-roll (approved):

car-roll:
strange...
i  went ahead and kicked off a new release

PR #188, opened 2022-07-20
Fix flakiness by synchronising the State

Pldi23:
Description:

Cloudbees CI reported a flake in 
[org.jenkinsci.plugins.workflow.support.steps.StageStepTest.serializability](https://github.com/jenkinsci/pipeline-stage-step-plugin/blob/master/src/test/java/org/jenkinsci/plugins/workflow/support/steps/StageStepTest.java#L173-L200)

Error Log:
[stderr.txt](https://github.com/jenkinsci/workflow-support-plugin/files/9150162/stderr.txt)

How to reproduce:
I wasn't managed to reproduce it locally without code changes, but if we add `TimeUnit.MILLISECONDS.sleep(200);` before this [line](https://github.com/jenkinsci/workflow-support-plugin/blob/master/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java#L110) the StageStepTest.serializability started to flake regularly.

Assumption:
My assumption that there is a deadlock possibility.

Solution:
As a solution I want to suggest to synchronize the State local variable (the same synchronisation mechanism is used [here](https://github.com/jenkinsci/workflow-support-plugin/blob/master/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java#L141)) 

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
@@ -104,14 +104,14 @@ public final class SemaphoreStep extends AbstractStepImpl implements Serializabl
     public static void success(String k, Object returnValue) {
         State s = State.get();
         synchronized (s) {
-            if (s.contexts.containsKey(k)) {
-                System.err.println("Unblocking " + k + " as success");
-                getContext(s, k).onSuccess(returnValue);
-            } else {
+            if (!s.contexts.containsKey(k)) {
                 System.err.println("Planning to unblock " + k + " as success");
                 s.returnValues.put(k, returnValue);
+                return;
             }
         }
+        System.err.println("Unblocking " + k + " as success");
+        getContext(s, k).onSuccess(returnValue);
     }
 
     /** Marks the step as having failed; or, if not yet started, makes it do so synchronously when started. */
@@ -122,14 +122,14 @@ public final class SemaphoreStep extends AbstractStepImpl implements Serializabl
     public static void failure(String k, Throwable error) {
         State s = State.get();
         synchronized (s) {
-            if (s.contexts.containsKey(k)) {
-                System.err.println("Unblocking " + k + " as failure");
-                getContext(s, k).onFailure(error);
-            } else {
+            if (!s.contexts.containsKey(k)) {
                 System.err.println("Planning to unblock " + k + " as failure");
                 s.errors.put(k, error);
+                return;
             }
         }
+        System.err.println("Unblocking " + k + " as failure");
+        getContext(s, k).onFailure(error);
     }
     
     public StepContext getContext() {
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -119,12 +121,14 @@ public void failure(Throwable error) {
 
     public static void failure(String k, Throwable error) {
         State s = State.get();
-        if (s.contexts.containsKey(k)) {
-            System.err.println("Unblocking " + k + " as failure");
-            getContext(s, k).onFailure(error);
-        } else {
-            System.err.println("Planning to unblock " + k + " as failure");
-            s.errors.put(k, error);
+        synchronized (s) {
```
Pldi23, line 124: I did not notice some tests that failed because of the race condition here, but since we are synchronizing the success method, then we need to synchronize here too.

jmdesprez (approved):

Pldi23:
Waiting for a review from maintainers

Pldi23:
@jglick could you please review

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -103,12 +103,14 @@ public void success(Object returnValue) {
...
-        if (s.contexts.containsKey(k)) {
-            System.err.println("Unblocking " + k + " as success");
-            getContext(s, k).onSuccess(returnValue);
-        } else {
-            System.err.println("Planning to unblock " + k + " as success");
-            s.returnValues.put(k, returnValue);
+        synchronized (s) {
+            if (s.contexts.containsKey(k)) {
+                System.err.println("Unblocking " + k + " as success");
+                getContext(s, k).onSuccess(returnValue);
```
jglick, line 109: I am somewhat concerned about holding a lock while calling `onSuccess` here. Is it effective to hold the lock on `State` only while accessing its fields? For example

```java
public static void success(String k, Object returnValue) {
    State s = State.get();
    synchronized (s) {
        if (!s.contexts.containsKey(k)) {
            System.err.println("Planning to unblock " + k + " as success");
            s.returnValues.put(k, returnValue);
            return;
        }
    }
    System.err.println("Unblocking " + k + " as success");
    getContext(s, k).onSuccess(returnValue);
}
```

jglick (commented):
IIUC to match https://github.com/jenkinsci/workflow-support-plugin/blob/a1c71061486b4ce3db98576f9435f7ca054b6aec/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java#L141-L148

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
Pldi23: Yeah that's even better.

jglick (approved):

PR #221, opened 2023-07-20
Remove usages of Prototype

basil:
To test this I ran the test suite against a version of Jenkins with Prototype removed. Before this PR the test suite failed; after this PR, the test suite passed.


jglick (approved):
Tested in `hpi:run` by configuring the Jenkins root URL, installing `pipeline-input-step`, creating a Pipeline

```groovy
input 'OK?'
```

and running it, then clicking the **Proceed** prompt from the standard log view.

PR #235, opened 2023-09-11
Fixing synchronization in `SemaphoreStep`

jglick:
#188 does not seem to have been complete; various `HashMap`-valued fields in `State` were accessed without consistent locking.

Motivated by frequent flakes in `SCMRetrieverTest.selfTestLibraries` (from `pipeline-groovy-lib`) in PCT. The test would time out waiting for one of the branch builds to complete; log messages _Blocking…_ indicated that the step had been started on the right string, but then later the log would show _Planning to unblock…_ rather than _Unblocking…_, as if `contexts` did not contain the string which had been added. I was able to reproduce such a failure locally after updating the core & BOM to newest 2.414.x and running the test about ten times in a row. My hypothesis is that unsynchronized access corrupted the `HashMap`.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
@@ -24,7 +24,6 @@
 
 package org.jenkinsci.plugins.workflow.test.steps;
 
-import com.google.inject.Inject;
 import hudson.Extension;
 import hudson.model.Run;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -96,47 +101,61 @@ private String k() {
...
     }
 
+    /** Marks the step as having successfully completed; or, if not yet started, makes it do so synchronously when started. */
     public static void success(String k, Object returnValue) {
         State s = State.get();
+        StepContext c;
         synchronized (s) {
             if (!s.contexts.containsKey(k)) {
-                System.err.println("Planning to unblock " + k + " as success");
+                LOGGER.info(() -> "Planning to unblock " + k + " as success");
```
jglick, line 116: Using `Logger` is better style and gives us timestamps too.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -147,29 +166,50 @@ public static void waitForStart(@NonNull String k, @CheckForNull Run<?,?> b) thr
                 if (b != null && !b.isBuilding()) {
                     throw new AssertionError(JenkinsRule.getLog(b));
                 }
-                s.wait(1000);
+                s.wait(100);
```
jglick, line 169: Might shave off a fraction of a second from tests to check this more promptly.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -36,34 +35,40 @@
...
 import org.jvnet.hudson.test.JenkinsRule;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 /**
  * Step that blocks until signaled.
- * Starts running and waits for {@link #success(Object)} or {@link #failure(Throwable)} to be called, if they have not been already.
+ * Starts running and waits for {@link #success(String, Object)} or {@link #failure(String, Throwable)} to be called, if they have not been already.
  */
-public final class SemaphoreStep extends AbstractStepImpl implements Serializable {
+public final class SemaphoreStep extends Step implements Serializable {
```
jglick, line 53: Cleaning up deprecations.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -147,29 +166,50 @@ public static void waitForStart(@NonNull String k, @CheckForNull Run<?,?> b) thr
...
-            if (s.returnValues.containsKey(k)) {
-                System.err.println("Immediately running " + k);
-                getContext().onSuccess(s.returnValues.get(k));
-            } else if (s.errors.containsKey(k)) {
-                System.err.println("Immediately failing " + k);
-                getContext().onFailure(s.errors.get(k));
+            Object returnValue = null;
+            Throwable error = null;
+            boolean success = false, failure = false, sync = true;
+            synchronized (s) {
```
jglick, line 192: Another point where synchronization was inadequate. Rather cumbersome idiom to avoid holding the lock while making Pipeline API calls, to avoid possible deadlocks.

julieheard (approved):

PR #252, opened 2024-02-07
`DefaultStepContext.get` should directly handle certain types

jglick:
We do not expect `CpsStepContext.doGet` to ever return non-null for these so why even ask? May solve test flakes introduced by https://github.com/jenkinsci/pipeline-input-step-plugin/pull/145.


dwnusbaum (approved):
This looks fine to me and we have considered it previously. I think it is worth running PCT against it first though to check for any unexpected consequences.

dwnusbaum (approved):

PR #258, opened 2024-03-28
Try to fix synchronization in `SemaphoreStep`

dwnusbaum:
Even after #235, I saw a test flake because a `semaphore` step hung forever even though `SempahoreStep.success` was called. The logs indicate an order of operations that show a bug in the existing synchronization:

```
  10.238 [id=1884]	INFO	o.j.p.w.t.s.SemaphoreStep$Execution#start: Blocking wait/1
  10.239 [id=1839]	INFO	o.j.p.w.test.steps.SemaphoreStep#success: Planning to unblock wait/1 as success
```

Here is my hypothesis as to what happened. We have two threads: Thread 1 calls `SemaphoreStep.start`. Thread 2 calls `SemaphoreStep.success`. They race for the `State` monitor. I think that thread 1 acquired the `State` monitor first and ran this entire block:

https://github.com/jenkinsci/workflow-support-plugin/blob/1e84e32f7b068c0d4713a7e18b2a68eb25048702/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java#L192-L200

Thread 2 then acquired the `State` monitor and ran this block:

https://github.com/jenkinsci/workflow-support-plugin/blob/1e84e32f7b068c0d4713a7e18b2a68eb25048702/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java#L136-L143

At this point, `s.contexts` does not contain `k`, so `Planning to unblock wait/1 as success` gets logged, and the monitor is released, but **thread 1 already checked `s.returnValues` and will never check it again**. Thread 1 in the meantime proceeded to this `synchronized` block and ran it once thread 2 released the monitor:

https://github.com/jenkinsci/workflow-support-plugin/blob/1e84e32f7b068c0d4713a7e18b2a68eb25048702/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java#L210-L212

At this point though, this does nothing. Thread 2 already ran `SemaphoreStep.success`, and the context did not exist, so it put a value in `State.returnValues`. Thread 1 had already checked `SemaphoreStep.returnValues` though, and will not check it again, so the step just hangs forever and the test times out.

CC @jgreffe

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
@@ -189,6 +189,7 @@ public final class SemaphoreStep extends Step implements Serializable {
             Object returnValue = null;
             Throwable error = null;
             boolean success = false, failure = false, sync = true;
+            String c = Jenkins.XSTREAM.toXML(getContext());
             synchronized (s) {
                 if (s.returnValues.containsKey(k)) {
                     success = true;
@@ -196,6 +197,8 @@ public final class SemaphoreStep extends Step implements Serializable {
                 } else if (s.errors.containsKey(k)) {
                     failure = true;
                     error = s.errors.get(k);
+                } else {
+                    s.contexts.put(k, c);
                 }
             }
             if (success) {
@@ -206,10 +209,6 @@ public final class SemaphoreStep extends Step implements Serializable {
                 getContext().onFailure(error);
             } else {
                 LOGGER.info(() -> "Blocking " + k);
-                String c = Jenkins.XSTREAM.toXML(getContext());
-                synchronized (s) {
-                    s.contexts.put(k, c);
-                }
                 sync = false;
             }
             synchronized (s) {


src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -189,13 +189,16 @@ public static class Execution extends AbstractStepExecutionImpl {
...
+            String c = Jenkins.XSTREAM.toXML(getContext());
             synchronized (s) {
                 if (s.returnValues.containsKey(k)) {
                     success = true;
                     returnValue = s.returnValues.get(k);
                 } else if (s.errors.containsKey(k)) {
                     failure = true;
                     error = s.errors.get(k);
+                } else {
+                    s.contexts.put(k, c);
```
dwnusbaum, line 201: This is the fix. Other threads now only ever see `State` in one of two modes: either `contexts` does not contain the relevant key, and `returnValues` and `errors` should be used and will be checked here, or `contexts` does contain the key, and the `StepContext` should be used instead. Previously, there was a case where `contexts` did not contain the key but `returnValues` and `errors` had already been checked.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -206,10 +209,6 @@ public static class Execution extends AbstractStepExecutionImpl {
                 getContext().onFailure(error);
             } else {
                 LOGGER.info(() -> "Blocking " + k);
-                String c = Jenkins.XSTREAM.toXML(getContext());
-                synchronized (s) {
-                    s.contexts.put(k, c);
-                }
                 sync = false;
```
dwnusbaum, line 212: FWIW I think there may be other pre-existing bugs here. What happens if another thread calls `success` or `failure` after the first `synchronized` block above but before this completes? Those threads will call `StepContext.onSuccess`/`StepContext.onFailure` before `StepExecution.start` has completed, which seems unusual, but maybe it's fine.

dwnusbaum: Maybe we should change the overall approach, drop support for the synchronous completion cases, and have the step just poll continuously on a background thread looking for results. It would be slower and less efficient, but it also seems less prone to complex synchronization issues.

jglick:
https://github.com/jenkinsci/support-core-plugin/pull/515#issuecomment-1973102400 ?

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
dwnusbaum: > Maybe we should change the overall approach

FWIW, I filed https://github.com/jenkinsci/workflow-support-plugin/pull/259 with an example of what that could look like.

jglick:
Perhaps it would be easier to reason about if instead of three maps there were just one, with values being a union of Blocked | Success<Object> | Error<Throwable>?

dwnusbaum:
> https://github.com/jenkinsci/support-core-plugin/pull/515#issuecomment-1973102400 ?

I saw the issue in a proprietary plugin, but indeed the logs for that test failure look the same:

```
   9.938 [id=63]	INFO	o.j.p.w.t.s.SemaphoreStep$Execution#start: Blocking wait/1
   9.939 [id=141]	INFO	o.j.p.w.test.steps.SemaphoreStep#success: Planning to unblock wait/1 as success
```

jglick:
Sorry, _four_ maps, I forgot about `started`.

dwnusbaum:
> Perhaps it would be easier to reason about if instead of [four] maps there were just one, with values being a union...

Yes, but this pattern is a bit awkward in Java. I can file another PR with that approach for comparison.

dwnusbaum:
See https://github.com/jenkinsci/workflow-support-plugin/pull/260.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
jglick: > Those threads will call `StepContext.onSuccess`/`StepContext.onFailure` before `StepExecution.start` has completed

That would be normal if `false` is returned. It would be weird if `start` later returns `true`, but I do not think it is harmful. TBH I do not recall the logic here in `workflow-cps` very well. Anyway this seems unlikely in a normal test, because it should call `wait` and then do some stuff before calling `success`.

jglick (approved):
I have a general preference for something like #260 as a refactoring, but OTOH this is fairly simple and looks like it may solve the practical issues, so you should go ahead and get _something_ tested & released.

jglick:
Feel free to release this PR if it seems to pass PCT. No need to waste days on this.

PR #259, opened 2024-03-28
Alternate approach to SemaphoreStep using Timer

dwnusbaum:
Alternative to https://github.com/jenkinsci/workflow-support-plugin/pull/258 that aims for simplicity over performance. More or less untested, and at the very least it changes some of the logging, so we'd probably want to run the PCT against this first if we want to try this approach.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
@@ -189,6 +189,7 @@ public final class SemaphoreStep extends Step implements Serializable {
             Object returnValue = null;
             Throwable error = null;
             boolean success = false, failure = false, sync = true;
+            String c = Jenkins.XSTREAM.toXML(getContext());
             synchronized (s) {
                 if (s.returnValues.containsKey(k)) {
                     success = true;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -134,29 +127,9 @@ public static void failure(String k, Throwable error) {
...
-    }
-
-    /** @deprecated should not be needed */
-    @Deprecated
-    public StepContext getContext() {
-        State s = State.get();
-        synchronized (s) {
-            return getContext(s, k());
-        }
-    }
```
dwnusbaum, line 155: Could be restored if needed, but I didn't see any callers in a quick search, and IDK how it would even be used in the first place. Deleting this though allows us to totally drop `State.contexts.`

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -185,10 +158,38 @@ public static class Execution extends AbstractStepExecutionImpl {
...
+
+        @Override
+        public void onResume() {
+            Timer.get().submit(this::checkStatus);
+        }
+
+        @Override public String getStatus() {
+            State s = State.get();
+            synchronized (s) {
+                return "waiting on " + k;
```
dwnusbaum, line 184: Message is less precise now, could be restored if needed.

jglick (commented):
This one makes me nervous; adding a background thread seems like it would just increase the chances of nondeterministic test failures.

PR #260, opened 2024-03-28
Move `SemaphoreStep` states into a single map to try to simplify the implementation

dwnusbaum:
See https://github.com/jenkinsci/workflow-support-plugin/pull/258#issuecomment-2026171643.

Personally, I feel like this is still pretty comparable to https://github.com/jenkinsci/workflow-support-plugin/pull/258 in terms of complexity. It seems a little better in terms of synchronization complexity, but the state transition complexity seems about the same or worse. The issue is that the four maps did not represent distinct non-overlapping states and so it is still not that clear what is going on in the new code. If we are purely going for simplicity, something like https://github.com/jenkinsci/workflow-support-plugin/pull/259 would be my vote, then #258, then this PR.

If we want to go with this approach, I think this PR would need a PCT run because it seems possible that is accidentally broke some typical use case.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
@@ -108,12 +108,15 @@ public final class SemaphoreStep extends Step implements Serializable {
         StepContext c;
         synchronized (s) {
             KeyState keyState = s.keyStates.get(k);
-            if (!(keyState instanceof StartedState)) {
+            if (keyState == null || keyState instanceof ImmediateSuccessState || keyState instanceof ImmediateFailureState) {
                 LOGGER.info(() -> "Planning to unblock " + k + " as success");
                 s.keyStates.put(k, new ImmediateSuccessState(returnValue));
                 return;
+            } else if (keyState instanceof FinishedState) {
+                throw new IllegalStateException(k + " already finished");
             }
             c = getContext(s, k);
+            s.keyStates.put(k, new FinishedState());
         }
         LOGGER.info(() -> "Unblocking " + k + " as success");
         c.onSuccess(returnValue);
@@ -131,12 +134,15 @@ public final class SemaphoreStep extends Step implements Serializable {
         StepContext c;
         synchronized (s) {
             Object keyState = s.keyStates.get(k);
-            if (!(keyState instanceof StartedState)) {
+            if (keyState == null || keyState instanceof ImmediateSuccessState || keyState instanceof ImmediateFailureState) {
                 LOGGER.info(() -> "Planning to unblock " + k + " as failure");
                 s.keyStates.put(k, new ImmediateFailureState(error));
                 return;
+            } else if (keyState instanceof FinishedState) {
+                throw new IllegalStateException(k + " already finished");
             }
             c = getContext(s, k);
+            s.keyStates.put(k, new FinishedState());
         }
         LOGGER.info(() -> "Unblocking " + k + " as failure");
         c.onFailure(error);
@@ -159,7 +165,8 @@ public final class SemaphoreStep extends Step implements Serializable {
     public static void waitForStart(@NonNull String k, @CheckForNull Run<?,?> b) throws IOException, InterruptedException {
         State s = State.get();
         synchronized (s) {
-            while (!(s.keyStates.get(k) instanceof StartedState)) {
+            KeyState keyState;
+            while (!(((keyState = s.keyStates.get(k)) instanceof StartedState) || keyState instanceof FinishedState)) {
                 if (b != null && !b.isBuilding()) {
                     throw new AssertionError(JenkinsRule.getLog(b));
                 }
@@ -192,13 +199,15 @@ public final class SemaphoreStep extends Step implements Serializable {
                 if (keyState instanceof ImmediateSuccessState) {
                     success = true;
                     returnValue = ((ImmediateSuccessState) keyState).returnValue;
+                    s.keyStates.put(k, new FinishedState());
                 } else if (keyState instanceof ImmediateFailureState) {
                     failure = true;
                     error = ((ImmediateFailureState) keyState).error;
+                    s.keyStates.put(k, new FinishedState());
                 } else if (keyState == null) {
                     s.keyStates.put(k, new StartedState(c));
                 } else {
-                    throw new IllegalStateException("Unable to start semaphore step " + k + " in state " + keyState);
+                    throw new IllegalStateException("Unable to start " + k + " in state " + keyState);
                 }
             }
             if (success) {
@@ -212,8 +221,6 @@ public final class SemaphoreStep extends Step implements Serializable {
                 sync = false;
             }
             synchronized (s) {
-                // Even if we completed immediately, set the state to StartedState so waitForSuccess knows to stop waiting.
-                s.keyStates.put(k, new StartedState(c));
                 s.notifyAll();
             }
             return sync;
@@ -222,7 +229,7 @@ public final class SemaphoreStep extends Step implements Serializable {
         @Override public void stop(Throwable cause) throws Exception {
             State s = State.get();
             synchronized (s) {
-                s.keyStates.remove(k);
+                s.keyStates.put(k, new FinishedState());
             }
             LOGGER.log(Level.INFO, cause, () -> "Stopping " + k);
             super.stop(cause);
@@ -231,7 +238,16 @@ public final class SemaphoreStep extends Step implements Serializable {
         @Override public String getStatus() {
             State s = State.get();
             synchronized (s) {
-                return s.keyStates.get(k) == null ? "waiting on " + k : "finished " + k ;
+                KeyState keyState = s.keyStates.get(k);
+                if (keyState instanceof ImmediateSuccessState) {
+                    return k + " will immediately succeed";
+                } else if (keyState instanceof ImmediateFailureState) {
+                    return k + " will immediately fail";
+                } else if (keyState instanceof FinishedState) {
+                    return "finished " + k;
+                } else {
+                    return "waiting on " + k;
+                }
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -134,12 +133,16 @@ public static void failure(String k, Throwable error) {
...
+            Object keyState = s.keyStates.get(k);
+            if (keyState == null) {
                 LOGGER.info(() -> "Planning to unblock " + k + " as failure");
-                s.errors.put(k, error);
+                s.keyStates.put(k, new ImmediateFailureState(error));
                 return;
+            } else if (keyState instanceof StartedState) {
+                c = getContext(s, k);
+            } else {
+                throw new IllegalStateException("Cannot mark " + k + " as failure from state " + keyState);
```
dwnusbaum, line 144: ditto

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -222,7 +228,7 @@ public static class Execution extends AbstractStepExecutionImpl {
         @Override public void stop(Throwable cause) throws Exception {
             State s = State.get();
             synchronized (s) {
-                s.contexts.remove(k);
+                s.keyStates.remove(k);
```
dwnusbaum, line 231: I am not sure that this behavior is the same. Would have to think about it some more.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -231,7 +237,7 @@ public static class Execution extends AbstractStepExecutionImpl {
         @Override public String getStatus() {
             State s = State.get();
             synchronized (s) {
-                return s.contexts.containsKey(k) ? "waiting on " + k : "finished " + k;
+                return s.keyStates.get(k) == null ? "waiting on " + k : "finished " + k ;
```
dwnusbaum, line 240: I think this behavior is the same as before, but I'm not 100% sure. Notably, it returns "waiting on" after `success` and `failure` have been called (the old code did too). To fix that, I think we'd need to add another state.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -206,14 +209,11 @@ public static class Execution extends AbstractStepExecutionImpl {
...
                 LOGGER.info(() -> "Blocking " + k);
-                String c = Jenkins.XSTREAM.toXML(getContext());
-                synchronized (s) {
-                    s.contexts.put(k, c);
-                }
                 sync = false;
             }
             synchronized (s) {
-                s.started.add(k);
+                // Even if we completed immediately, set the state to StartedState so waitForSuccess knows to stop waiting.
```
jglick, line 215: Why? I would expect the wait to just wait for the key to be present, with any value.

Also did you mean
```suggestion
                // Even if we completed immediately, set the state to StartedState so waitForStart knows to stop waiting.
```
?

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
jglick: Seems safer to switch to the failure state. Should be idempotent right? There is no real reason to remove entries from the map.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
jglick: Does not look right. As above, just do not remove entries from the map. Then this will say it is finished when it is.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -255,5 +255,38 @@ public static class Execution extends AbstractStepExecutionImpl {
 
     }
 
+    // Marker interface just for clarity.
+    private interface KeyState { }
+
+    /**
+     * The step has not yet started, but will succeed immediately when it does start.
+     */
+    private static class ImmediateSuccessState implements KeyState {
```
jglick, line 264: Clearer IMO to just call it a success state. The timing of that success is a matter for the code moving it around in the map.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -255,5 +255,38 @@ public static class Execution extends AbstractStepExecutionImpl {
...
+     */
+    private static class ImmediateFailureState implements KeyState {
+        private final Throwable error;
+        private ImmediateFailureState(Throwable error) {
+            this.error = error;
+        }
+    }
+
+    /**
+     * The step has at least started, and may have finished.
```
jglick, line 282: > may have finished

??

jglick (commented):
Not sure the details of the states are right, but the overall structure seems clearer to me.

jglick:
> this PR would need a PCT run

We should do a PCT run for any proposed change to this class. (`jenkinsci/bom` and/or @cloudbees internal)

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
dwnusbaum: Because otherwise if you call `SemaphoreStep.success` or `SemaphoreStep.failure` and then `SemaphoreStep.waitForStart` _without even running a `Pipeline` that uses the `semaphore` step_, `waitForStart` will complete immediately.

dwnusbaum: Basically, to really get this right, there needs to be at least 5 states, and we need to add a few more transitions that currently do not exist:
* No entry in the map 
* ImmediateFailure (the step will fail, if/when it starts)
* ImmediateSuccess (the step will succeed, if/when it starts)
* Started/Waiting/Blocked (the step has started, and has not yet completed)
* Finished (the step has finished, with some result) (does not currently exist, we'd need to add various transitions to this state)

And I do not think that we can introduce the `Finished` state without also introducing various behavior changes in edge cases that might affect existing tests. We'd definitely need a PCT run.

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
dwnusbaum: Adjusted in [d05eafd](https://github.com/jenkinsci/workflow-support-plugin/pull/260/commits/d05eafd27820d48db69890d04658053da4887743).

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
dwnusbaum: Adjusted in [d05eafd](https://github.com/jenkinsci/workflow-support-plugin/pull/260/commits/d05eafd27820d48db69890d04658053da4887743)

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
dwnusbaum: I think it is important to be able to distinguish "the step will succeed, when/if it starts" from "the step ran and finished". Maybe not for the common case, but for edge cases it seems important, and the old code was able to distinguish these states. Either way, I adjusted things in [d05eafd](https://github.com/jenkinsci/workflow-support-plugin/pull/260/commits/d05eafd27820d48db69890d04658053da4887743).

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -108,12 +108,15 @@ public static void success(String k, Object returnValue) {
         StepContext c;
         synchronized (s) {
             KeyState keyState = s.keyStates.get(k);
-            if (!(keyState instanceof StartedState)) {
+            if (keyState == null || keyState instanceof WillSucceedState || keyState instanceof WillFailState) {
```
jglick, line 111: Maybe introduce a common supertype `WillFinishState`.

jglick (approved):

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -112,12 +107,16 @@ public static void success(String k, Object returnValue) {
...
         synchronized (s) {
-            if (!s.contexts.containsKey(k)) {
+            KeyState keyState = s.keyStates.get(k);
+            if (keyState == null || keyState instanceof WillSucceedState || keyState instanceof WillFailState) {
                 LOGGER.info(() -> "Planning to unblock " + k + " as success");
-                s.returnValues.put(k, returnValue);
+                s.keyStates.put(k, new WillSucceedState(returnValue));
                 return;
+            } else if (keyState instanceof FinishedState) {
+                throw new IllegalStateException(k + " already finished");
```
dwnusbaum, line 116: This breaks tests. In particular at least `workflow-step-api / org.jenkinsci.plugins.workflow.steps.GeneralNonBlockingStepExecutionTest.stop` and `workflow-cps / org.jenkinsci.plugins.workflow.cps.CpsFlowExecutionTest.stepsAreStoppedWhenCpsVmExecutorServiceHandlesUncaughtException`. I'll take a look.

dwnusbaum: `GeneralNonBlockingStepExecutionTest.stop` I think just has a redundant call to `SemaphoreStep.success` that could be removed.

`CpsFlowExecutionTest.stepsAreStoppedWhenCpsVmExecutorServiceHandlesUncaughtException` is a little unique, basically the test was using `semaphore` as a way to call `StepContext.onSuccess` after a Pipeline died. To preserve that test's behavior without changes, we'd have to make `FinishedState` keep the serialized context around. Probably clearer to update that test to use a custom step instead.

dwnusbaum:
I have no plans to finish this up, so I am closing it; there are two tests mentioned above in other plugins would require changes either here or to the tests depending on the desired behavior.

PR #263, opened 2024-04-29
Use lists instead of arrays for `FlowNode` actions to make deserialization more robust

dwnusbaum:
I did not realize this, but if you try to deserialize a Pipeline that has a `FlowNode` with an action from a plugin that is disabled or no longer installed, the `FlowNode` itself will fail to be deserialized, leaving the flow graph corrupted. It seems like it would be preferential to have `RobustCollectionConverter` work for `FlowNode` actions, so that only the problematic action itself fails to be deserialized, but the `FlowNode` itself and all other actions are preserved. 

The reason for the current behavior is that right now, the fields used to serialize `FlowNode` actions are arrays, and core's [`RobustCollectionConverter`](https://github.com/jenkinsci/jenkins/blob/cabf1a9cece07581420da2ecb9bf9c6d679c2b8d/core/src/main/java/hudson/util/RobustCollectionConverter.java) has no equivalent for arrays (fair enough I guess).

I created some tests to demonstrate the original behavior, but then I realized that the serial format of lists and arrays as far as XStream is concerned actually seems to be interchangeable (maybe not true in general?), so I figured I might as well file a patch instead of just tests. Besides potential concerns with backwards compatibility, I am not sure if there is any specific reason that arrays were used originally. Maybe to reduce memory usage or something like that?

For now I'm just filing as a draft because I'm not sure if this is safe or if there were any particular reasons we were using arrays previously.

### Testing done

Only tested briefly against this plugin using the new and existing automated tests.

```[tasklist]
### Submitter checklist
```


src/test/resources/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorageTest/actionDeserializationShouldBeRobust/.lastStarted
dwnusbaum, line 1: Probably various files here could be deleted, and in general I prefer to use zip files to avoid cluttering GitHub search results, but for now at least having the bare files as test resources makes it easy to see what is being tested.

src/test/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorageTest.java
```diff
@@ -58,4 +68,37 @@ public final class BulkFlowNodeStorageTest {
...
+        SemaphoreStep.waitForStart("wait/1", b);
+        ((CpsFlowExecution) b.getExecution()).getStorage().flush();
+        Thread.sleep(30*1000);
+        */
+        var p = r.jenkins.getItemByFullName("test0", WorkflowJob.class);
+        var b = p.getLastBuild();
+        var stageBodyStartNode = (StepStartNode) b.getExecution().getNode("4");
+        assertThat(stageBodyStartNode, not(nullValue()));
+        var label = stageBodyStartNode.getPersistentAction(LabelAction.class);
+        assertThat(label.getDisplayName(), equalTo("test"));
```
dwnusbaum, line 90: Without the patch, the whole Pipeline itself fails to load because the active head refers to a node with the action from the class that is no longer present.

src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -203,7 +204,7 @@ public void saveActions(@NonNull FlowNode node, @NonNull List<Action> actions) t
         if (t != null) {
             t.node = node;
             List<Action> act = node.getActions();
-            t.actions = act.toArray(new Action[act.size()]);
+            t.actions = new ArrayList<>(actions);
```
dwnusbaum, line 207:
```suggestion
            t.actions = new ArrayList<>(act);
```

Dohbedoh (approved):

src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java
```diff
@@ -234,6 +234,23 @@ public static class Execution extends AbstractStepExecutionImpl {
             }
         }
 
+        @Override public void onResume() {
```
dwnusbaum, line 237: I added an `onResume` implementation for `semaphore` so that it can be used with `@LocalData`. I don't think it should make a difference for tests that use `JenkinsSessionRule`, although maybe we should guard this logic and do nothing if the state is not empty. Let me see if we have decent test coverage for `semaphore` here.

jglick: I think I have generally used a real step in conjunction with `@LocalData`; feels a bit uneasy to serialize a `SemaphoreStep.Execution` in persistent test data.

dwnusbaum: Fair enough. I switched to `sleep` in [b5e199f](https://github.com/jenkinsci/workflow-support-plugin/pull/263/commits/b5e199fde4eb418111d162a5a28402d05ff4caa2).

jtnord (approved):

jglick (approved):

PR #278, opened 2024-09-17
Add a timeout for pickle resolution to prevent indefinitely stuck builds

dwnusbaum:
Related to https://github.com/jenkinsci/workflow-api-plugin/pull/357 and https://github.com/jenkinsci/workflow-api-plugin/pull/347. Even though the fix there should prevent stuck builds from leaking memory of other builds, it is still worth trying to prevent builds from getting stuck in the first place.

This PR adds a hard timeout to pickle resolution. I do not care much about the exact timeout, I just want to pick a value that doesn't break pickle resolution in non-pathological cases. The case described in https://github.com/jenkinsci/workflow-api-plugin/pull/357 involved a build repeatedly resuming and trying to resolve a `FilePathPickle` that was never going to come back for around a month.

If preferred this could be done in `CpsFlowExecution.loadProgramAsync` or `TryRepeatedly` instead.

### Submitter checklist
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/PickleResolver.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/PickleResolver.java
@@ -24,16 +24,22 @@
 
 package org.jenkinsci.plugins.workflow.support.pickles.serialization;
 
-import org.jenkinsci.plugins.workflow.pickles.Pickle;
-import org.jenkinsci.plugins.workflow.support.concurrent.Futures;
 import com.google.common.base.Function;
+import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import org.jboss.marshalling.ObjectResolver;
-
+import com.google.common.util.concurrent.MoreExecutors;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/pickles/serialization/PickleResolver.java
```diff
@@ -24,23 +24,38 @@
...
  */
 public class PickleResolver implements ObjectResolver {
+
+    /**
+     * Pickle resolution will fail automatically after this many seconds.
+     * <p>This is intended to prevent Pipeline builds from hanging forever in unusual cases.
+     */
+    @Restricted(NoExternalUse.class)
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "Non-final for modification from script console")
+    public static long RESOLUTION_TIMEOUT_SECONDS = SystemProperties.getLong(PickleResolver.class + ".RESOLUTION_TIMEOUT_SECONDS", TimeUnit.HOURS.toSeconds(1));
```
dwnusbaum, line 57: Not sure what value is best here. One hour resolving pickles seemed roughly ok to me. I think we don't really expect `TryRepeatedly` to be used normally anyway now that `ExecutorStepDynamicContext` exists, but there are a couple of [other usages](https://github.com/search?type=code&q=+owner%3Ajenkinsci+tryrepeatedly).

PR #345, opened 2025-09-18
WithThreadName: add a way to disable with a JVM property [JENKINS-75967]

jimklimov:
Grasping at straws for https://issues.jenkins.io/browse/JENKINS-75967 to rule out whether this native call is really as costly on that server as clues seem to indicate, or the problem is elsewhere.

This PR adds support for a JVM property `DISABLE_WithThreadName` that should get evaluated once per runtime (as a `Boolean.valueOf()`, so `true` vs. anything else including `null`) to skip out of both methods in this class early.

### Testing done

TBD manually with incremental built artifact. Maybe unit tests later.

### Submitter checklist


jimklimov:
Updated to log the disabled runs, to help gauge on the impacted system when and how often this code is called.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/WithThreadName.java
```diff
@@ -24,20 +24,44 @@
...
+    private static final Logger LOGGER = Logger.getLogger(WithThreadName.class.getName());
+
     /**
      * Sets the current thread’s name.
      * @param suffix text to append to the original name
      */
     public WithThreadName(String suffix) {
+        if (disabled) {
+            original = null;
+            LOGGER.log(Level.FINE, "SKIP: Neutered WithThreadName(\"{0}\")", suffix);
```
jglick, line 61: BTW the `fine(Supplier<String>)` overload is generally more legible.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/WithThreadName.java
```diff
@@ -24,20 +24,44 @@
...
+     */
     private final String original;
 
+    /** Optional toggle via JVM properties to skip work here,
+     *  and forfeit easy debugging, e.g. on systems where
+     *  java.lang.Thread.setNativeName(Native Method) aka
+     *  JVM_SetNativeThreadName() and further platform
+     *  specific implementation takes inexplicably long.
+     */
+    private final static Boolean disabled = Boolean.valueOf(System.getProperty("DISABLE_WithThreadName"));
```
jglick, line 49: `SystemProperties.getBoolean`

jimklimov:
@jglick : thanks for the review and suggestions, will try to implement those.

> Maybe just disable it (unconditionally, or by default) on OpenIndiana, since it sounds like a problem in the JVM port or in the OS itself?

I am not certain if that is a general platform problem, or some mis-alignment of stars on that specific controller. So the path of least surprise is to keep doing what was done before unless configured otherwise.

But as noted on Gitter, in logs of the massively-parallel job I see it do 100K entries (two per change/revert thread name) in 20 seconds, so maybe the root cause of the lags is elsewhere (e.g. someone too eagerly doing that in the first place, and causing tons of JNA calls to pthreads or wherever).

jimklimov:
Well, my uneducated guess is that this renaming generally was added to help some troubleshooting efforts, in which case having the info available seems useful regardless of intensity. It may be that this should have been disabled by default in the first place, and only enabled where/when debugging Jenkins internal workflows (via java props or somehow on the flight)?..

IMHO such change is not something for this PR to tackle, but better educated follow-up changes quite could.

jglick:
> this renaming generally was added to help some troubleshooting efforts

Yes, to help make sense of thread dumps.

> only enabled where/when debugging Jenkins internal workflows (via java props or somehow on the flight)

By the time you need this, that would be too late—it is intended to help debug things in a production controller.

src/main/java/org/jenkinsci/plugins/workflow/support/concurrent/WithThreadName.java
```diff
@@ -24,20 +24,46 @@
...
+     */
     private final String original;
 
+    /** Optional toggle via JVM properties to skip work here,
+     *  and forfeit easy debugging, e.g. on systems where
+     *  java.lang.Thread.setNativeName(Native Method) aka
+     *  JVM_SetNativeThreadName() and further platform
+     *  specific implementation takes inexplicably long.
+     */
+    private final static Boolean disabled = SystemProperties.getBoolean("DISABLE_WithThreadName");
```
jglick, line 51:
```suggestion
    private final static boolean enabled = SystemProperties.getBoolean(WithThreadName.class.getName() + ",enabled", true);
```
and then use `!enabled` elsewhere

jimklimov: Fair enough :)

jimklimov: Hm, is a comma okay in property name, or should that have been an underscore etc.?

With your proposal, would it go like `java -DWithThreadName,enabled=false ...` where I want it OFF?

jimklimov:
Something did not go right with testing of latest iteration, maybe a typo in JVM setting. Controller lags again :\

jimklimov:
Retrying test with a full class name in option:
```
-Dorg.jenkinsci.plugins.workflow.support.concurrent.WithThreadName.enabled=false
```

jimklimov:
@jglick : does this look good enough to merge now? :)

README.md
```diff
@@ -7,3 +7,30 @@ This plugin provides APIs that are used by core Pipeline plugins for features su
...
+a context with this feature.
+
+In some use-cases, e.g. with a generated `parallel` pipeline with hundreds of
+stages then waiting in a Jenkins queue for considerable time, this can amount
+to tens of thousands of native (not very efficient on some systems) calls per
+second just for this troubleshooting aid, and can cause considerable slow-down
+of the Jenkins controller.
+
+If you suspect your controller's performance is impacted by this, please
+configure a JVM property or context setting
```
jglick: Fine though I would still rather just see this done automatically than offer yet another thing for users to think about.

jimklimov: ESCOPECREEP :) That could be a next feature, if someone takes on it, and designs in a way that is easy to use and/or least-surprise for existing users.

If all this logging was introduced recently (I don't know) then the least-surprise could in fact be having that disabled by default and for the few developers/troubleshooters relying on nicer thread names to enable it somehow. At least now it is using not directly the JVM option but a Jenkins context reader that taps into that among other sources, thanks to you. Maybe there's a way to flip the switch while live?

PR #348, opened 2025-09-30
Forcibly load static XStream instances during Jenkins startup to prevent a one-build memory leak

dwnusbaum:
This PR tries to address https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/199#discussion_r2377121804 by forcing `BulkFlowNodeStorage.XSTREAM` and `SimpleFlowNodeStorage.XSTREAM` to be loaded during startup via `@Initializer`.

I think we only really need this for `BulkFlowNodeStorage` due to https://github.com/jenkinsci/workflow-cps-plugin/pull/807, since with that PR, if you are using `SimpleXStreamFlowNodeStorage` (the default), the first build that completes may be the first thing to load `BulkFlowNodeStorage`, causing the one-build memory leak observed in tests in `pipeline-groovy-lib`.

I added an initializer for `SimpleFlowNodeStorage.XSTREAM` at the same time mostly just for uniformity. 

### Testing done

I tested this against https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/199 with the code referenced in https://github.com/jenkinsci/pipeline-groovy-lib-plugin/pull/199#discussion_r2377121804 removed, and the test passed.

### Submitter checklist
- [x] Link to relevant pull requests, esp. upstream and downstream changes

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
@@ -36,6 +36,8 @@ import org.jenkinsci.plugins.workflow.support.PipelineIOUtils;
 
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import hudson.init.InitMilestone;
+import hudson.init.Initializer;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.Field;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/storage/BulkFlowNodeStorage.java
```diff
@@ -243,6 +245,12 @@ private void storeActions() {  // We've already loaded the actions, may as well
         }
     }
 
+    @Initializer(after = InitMilestone.EXTENSIONS_AUGMENTED)
```
dwnusbaum, line 248: I need to check when `ItemListener.resume` is called and make sure we complete `before` that.

jglick (approved):

PR #360, opened 2025-12-04
Mute successful rows, highlight failed and unstable rows in the Pipeline Steps table

daniel-beck:
I don't use more advanced Pipeline visualization, and the table wasn't really usable in larger Pipelines. So I created https://plugins.jenkins.io/better-pipeline-flowgraph-table/ -- and I remember I had a reason to do so (https://github.com/jenkins-infra/repository-permissions-updater/issues/4013#issuecomment-2237453696), but I don't remember what it was specifically.

IMO this is better built into the default plugin, so it works for everyone, and that's what this PR does.

This applies basically the same style as https://plugins.jenkins.io/better-pipeline-flowgraph-table/ does. If accepted, I would deprecate that plugin.

### Testing done

Manually run.

<details><summary>Screenshots</summary>

Screenshot 2025-12-04 at 20 07 44
Screenshot 2025-12-04 at 20 07 38
Screenshot 2025-12-04 at 20 07 28
Screenshot 2025-12-04 at 20 07 21
Screenshot 2025-12-04 at 20 06 57
Screenshot 2025-12-04 at 20 02 58

</details>

### Submitter checklist

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable/ajax.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable/ajax.jelly
@@ -45,7 +45,7 @@
 
         <j:forEach var="row" items="${it.rows}">
           <j:set var="node" value="${row.node}"/>
-          <tr class="workflow-support-row-status-${node.iconColor} workflow-support-row-executed-${row.isExecuted()}"> <!-- Muted if not executed -->
+          <tr class="workflow-support-row-status-${node.iconColor} workflow-support-row-executed-${row.executed}"> <!-- Muted if not executed -->
             <!-- tooltip is for now debugging only -->
             <j:set var="exec_state" value=""/>
             <j:if test="${!row.executed}"><j:set var="exec_state" value=" (${%not executed})"/></j:if>
--- a/src/main/resources/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable/ajax.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable/ajax.jelly
@@ -28,6 +28,9 @@
 -->
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:l="/lib/layout">
   <l:ajax>
+    <j:if test="${!ajax}">
+      <st:adjunct includes="org.jenkinsci.plugins.workflow.support.visualization.table.FlowGraphTable.adjunct"/>
+    </j:if>
     <div id="nodeGraph">
       <table class="jenkins-table">
         <j:set var="columns" value="${it.columns}"/>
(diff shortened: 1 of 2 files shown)


jglick:
> I haven't found out yet what I need to do for `row#isExecuted` to be false

I think you found out of band that this is e.g. skipping a Declarative stage?

src/main/resources/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable/ajax.jelly
```diff
@@ -42,7 +45,7 @@
 
         <j:forEach var="row" items="${it.rows}">
           <j:set var="node" value="${row.node}"/>
-          <tr style="opacity: ${row.isExecuted()?'1.0':'0.3'}"> <!-- Muted if not executed -->
+          <tr class="workflow-support-row-status-${node.iconColor} workflow-support-row-executed-${row.isExecuted()}"> <!-- Muted if not executed -->
```
jglick: or for consistency
```suggestion
          <tr class="workflow-support-row-status-${node.iconColor} workflow-support-row-executed-${row.executed}"> <!-- Muted if not executed -->
```

daniel-beck:
> I think you found out of band that this is e.g. skipping a Declarative stage?

Yes, "Restart from Stage" does it.

Here's what it looks like in one of the themes:

> Screenshot 2025-12-04 at 21 23 52

(But TBH that looks like a bug, why is it just the one row?)

src/main/resources/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable/ajax.jelly
```diff
@@ -42,7 +45,7 @@
 
         <j:forEach var="row" items="${it.rows}">
           <j:set var="node" value="${row.node}"/>
-          <tr style="opacity: ${row.isExecuted()?'1.0':'0.3'}"> <!-- Muted if not executed -->
```
jglick, line 45: amending https://github.com/jenkinsci/pipeline-plugin/pull/333 FTR

src/main/resources/org/jenkinsci/plugins/workflow/support/visualization/table/FlowGraphTable/ajax.jelly
```diff
@@ -28,6 +28,9 @@
 -->
 <j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:l="/lib/layout">
   <l:ajax>
+    <j:if test="${!ajax}">
```
jglick, line 31: Really belongs in https://github.com/jenkinsci/workflow-job-plugin/blob/130687fcfd2eb546728dfdb1ac6666f5af174a91/src/main/resources/org/jenkinsci/plugins/workflow/job/views/FlowGraphTableAction/index.jelly#L30 but that would be awkward.

jglick (approved):

jglick:
> that looks like a bug, why is it just the one row?

Because the action would only be applied to the block node that was skipped, I guess, not every node inside it. More sophisticated visualizations would handle this at a higher level, not by displaying every `FlowNode`.

daniel-beck:
> > that looks like a bug, why is it just the one row?
> 
> Because the action would only be applied to the block node that was skipped, I guess, not every node inside it. More sophisticated visualizations would handle this at a higher level, not by displaying every `FlowNode`.

Right, but that means `row#isExecuted()` (or `NotExecutedNodeAction#isExecuted`) is too simple, and should also check ancestors for presence of the action.

jglick:
Exactly.

PR #432, opened 2026-09-21
Deflaking `SimpleXStreamStorageTest.actionDeserializationShouldBeRobust` on Windows

jglick:
https://github.com/jenkinsci/workflow-support-plugin/pull/431/checks?check_run_id=106222304057 (#431) flaked `…\jobs\test0\builds\1\log: The process cannot access the file because it is being used by another process`; not sure if this would fix it or not (cannot reproduce locally on Windows 10).


src/test/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamStorageTest.java
```diff
@@ -106,6 +106,7 @@ public void testDeferWriteAndFlush() throws Exception {
         assertThat(stageBodyStartNode, not(nullValue()));
         var label = stageBodyStartNode.getPersistentAction(LabelAction.class);
         assertThat(label.getDisplayName(), equalTo("test"));
+        b.delete();
```
jtnord, line 109: does this kill in progress builds?

the log looked interesting.

```
test0 #1 was finalized but never started; assuming it was started earlier using @LocalData
   3.458 [id=73]	WARNING	o.j.h.t.RemainingActivityListener#onTearDown: test0 #1 still seems to be running, which could break deletion of log files or metadata
   3.463 [id=73]	INFO	hudson.lifecycle.Lifecycle#onStatusUpdate: Stopping Jenkins
   3.480 [id=73]	INFO	hudson.lifecycle.Lifecycle#onStatusUpdate: Jenkins stopped
   3.480    3.553 [id=73]	INFO	o.j.h.t.TemporaryDirectoryAllocator#dispose: deleting Z:\agent\workspace\s_workflow-support-plugin_PR-431\target\tmp\j h3716940680001526087
[test0 #1] Started
   3.559 [test0 #1] [Pipeline] Start of Pipeline
   3.564 [test0 #1] [Pipeline] stage
   3.567 [test0 #1] [Pipeline] { (test)
   3.570 [test0 #1] [Pipeline] sleep
   3.572 [test0 #1] Sleeping for 2 min 0 sec
   3.574 [test0 #1] Resuming build at Mon Sep 21 05:15:05 UTC 2026 after Jenkins restart
   3.576
```

ie the build is somehow started after jenkins starts stopping;

either that or the build watcher is running too slowly, which would also cause issues with deletion?

jtnord: right so BuildWather is a `ClassRule` and the `JenkinsRule` is a regular rule.  
The buildwatcher has its own threads performing the copy, and its not terminated along with Jenkins.

TL;DR don't mix a ClassRule based BuildWatcher with a regular Rule based JenkinsRule?  (and ensure the ordering of them so that BuildWatcher gets cleaned up first?)

jtnord (approved):
not convinced this is a fix.  The buildwatcher should really be shutdown but :shrug:

src/test/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamStorageTest.java
jglick: The build is resumed when `JenkinsRule` starts the controller. It is completed during the test.

> don't mix a ClassRule based BuildWatcher with a regular Rule based JenkinsRule?

We do so in hundreds of places. That is the normal idiom for Pipeline tests. It checks for a completed build.

Without knowing how to reproduce we can only speculate what process was accessing the `log` file. Perhaps `BuildWatcher` but perhaps something unrelated.

jtnord: > The build is resumed when `JenkinsRule` starts the controller. It is completed during the test.
> 
> > don't mix a ClassRule based BuildWatcher with a regular Rule based JenkinsRule?
> 
> We do so in hundreds of places. That is the normal idiom for Pipeline tests. It checks for a completed build.
> 
> Without knowing how to reproduce we can only speculate what process was accessing the `log` file. Perhaps `BuildWatcher` but perhaps something unrelated.

Right but the build logger logs **are** output after Jenkins has shutdown which should not happen!  (ie this is your smoking gun!)
we can speculate as to why this occured - but I think we can safely say the root cause is this.

jglick: Hmm https://ci.jenkins.io/job/Plugins/job/workflow-support-plugin/job/PR-431/1/testReport/junit/org.jenkinsci.plugins.workflow.support.storage/SimpleXStreamStorageTest/windows_17___Build__windows_17____actionDeserializationShouldBeRobust/ does in fact include

```
WARNING	o.j.h.t.RemainingActivityListener#onTearDown: test0 #1 still seems to be running, which could break deletion of log files or metadata
```

which I somehow missed before. That is not `BuildWatcher`; that is the build actually still running, despite https://github.com/jenkinsci/workflow-support-plugin/blob/4a0ca2bc0b64b5ddf327de28d68aa7a7b3796ef7/src/test/java/org/jenkinsci/plugins/workflow/support/storage/SimpleXStreamStorageTest.java#L104

jglick: https://github.com/jenkinsci/jenkins-test-harness/issues/672#issuecomment-1785165418 There are a lot of subtle race conditions in build completion that could be cleaned up with some sustained attention, which I have never found the time for.

